PR #49, opened 2014-12-17
Schema relative rewrite

### LINKED ISSUE #48: .htaccess: enabling "forcing https" and "suppressing the www." does not remove subdomain
In the `.htaccess` If I enable:

* The forcing of https
* redirection for www. prefixed urls to urls without the prefix

The visitor is not redirected from `www.example.com` to `https://example.com`, as I would have expected. Instead, the user is redirected to `https://www.example.com` (so the prefix isn't removed).

I am not an expert in htaccess url rewriting, but it seems like the subdomain is only removed for http connections and it redirects to the unprefixed, but unsecured version of the url (so it doesn't do anything for https connections). Wouldn't it be better to rewrite secure connections as well? Just wondering if this is an oversight or by design.

```

# ----------------------------------------------------------------------
# | Forcing `https://`                                                 |
# ----------------------------------------------------------------------

# Redirect from the `http://` to the `https://` version of the URL.
# https://wiki.apache.org/httpd/RewriteHTTPToHTTPS

<IfModule mod_rewrite.c>
   RewriteEngine On
   RewriteCond %{HTTPS} !=on
   RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
</IfModule>

# ----------------------------------------------------------------------
# | Suppressing / Forcing the `www.` at the beginning of URLs          |
# ----------------------------------------------------------------------

# The same content should never be available under two different
# URLs, especially not with and without `www.` at the beginning.
# This can cause SEO problems (duplicate content), and therefore,
# you should choose one of the alternatives and redirect the other
# one.
#
# By default `Option 1` (no `www.`) is activated.
# http://no-www.org/faq.php?q=class_b
#
# If you would prefer to use `Option 2`, just comment out all the
# lines from `Option 1` and uncomment the ones from `Option 2`.
#
# (!) NEVER USE BOTH RULES AT THE SAME TIME!

# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

# Option 1: rewrite www.example.com &#8594; example.com

<IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
    RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]
    RewriteRule ^ http://%1%{REQUEST_URI} [R=301,L]
</IfModule>

```

Full .htaccess that led to this behaviour [here](https://gist.github.com/ismay/f0ef0c7801b83c98e0f9)

ismay:
Will suppress www. at the beginning of urls, regardless of `http:` or `https:`, instead of only rewriting urls with `http:`.

Fixes #48

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -335,6 +335,9 @@ AddDefaultCharset utf-8
 #     the rewrite engine.
 #
 #     https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriteoptions
+#
+# (6) Set %{ENV:PROTO} variable, to allow rewrites to redirect with the
+#     appropriate schema automatically (http or https).
 
 <IfModule mod_rewrite.c>
 
@@ -352,6 +355,12 @@ AddDefaultCharset utf-8
 
     # (5)
     # RewriteOptions <options>
+
+    # (6)
+    RewriteCond %{HTTPS} =on
+    RewriteRule ^ - [env=proto:https]
+    RewriteCond %{HTTPS} !=on
+    RewriteRule ^ - [env=proto:http]
 
 </IfModule>
 
@@ -394,7 +403,7 @@ AddDefaultCharset utf-8
     RewriteEngine On
     RewriteCond %{HTTPS} !=on
     RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]
-    RewriteRule ^ http://%1%{REQUEST_URI} [R=301,L]
+    RewriteRule ^ %{ENV:PROTO}://%1%{REQUEST_URI} [R=301,L]
 </IfModule>
 
 # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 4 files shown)


ismay:
@AD7six Ok, I've adapted the pull request to fit your suggestions. Wasn't sure whether I should leave out the `RewriteCond %{HTTPS} !=on` condition in the `rewrite_www.conf`. Should that be removed?

Other than that I think it's good to go. Let me know what you think!

AD7six:
Awesome =).

The https-on test should be removed, but since that's a change in functionality I'd rather not create a new major release just for that so it can stay as it is in the mean time.

Thanks for your contribution!

ismay:
No problem, happy to help!

> The https-on test should be removed, but since that's a change in functionality I'd rather not create a new major release just for that so it can stay as it is in the mean time.

I can create an issue to add to the 3.0.0 milestone, as a reminder. Might be useful right?

AD7six:
Done =)

ismay:
Ok cool! :+1:

PR #51, opened 2015-01-05
Use IncludesNOEXEC instead of just Includes option

andrewmkhoury:
For security purposes it would be best to use IncludesNOEXEC option instead of just Includes.

See here:
http://httpd.apache.org/docs/current/misc/security_tips.html#ssi

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -903,14 +903,14 @@ FileETag None
 
 # <IfModule mod_include.c>
 #     <FilesMatch "\.combined\.js$">
-#         Options +Includes
+#         Options +IncludesNOEXEC
 #         AddOutputFilterByType INCLUDES application/javascript \
 #                                        application/x-javascript \
 #                                        text/javascript
 #         SetOutputFilter INCLUDES
 #     </FilesMatch>
 #     <FilesMatch "\.combined\.css$">
-#         Options +Includes
+#         Options +IncludesNOEXEC
 #         AddOutputFilterByType INCLUDES text/css
 #         SetOutputFilter INCLUDES
 #     </FilesMatch>
--- a/src/web_performance/file_concatenation.conf
+++ b/src/web_performance/file_concatenation.conf
@@ -17,14 +17,14 @@
 
 <IfModule mod_include.c>
     <FilesMatch "\.combined\.js$">
-        Options +Includes
+        Options +IncludesNOEXEC
         AddOutputFilterByType INCLUDES application/javascript \
                                        application/x-javascript \
                                        text/javascript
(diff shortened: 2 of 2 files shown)


AD7six:
Seems appropriate, I doubt anyone is using exec. Can you please apply your change also to [the source file](https://github.com/h5bp/server-configs-apache/blob/master/src/web_performance/file_concatenation.conf)? Otherwise the next time the dist file is generated your change is lost.

andrewmkhoury:
Hi Andy,
Ok I made the same change on the source file you linked to
https://github.com/h5bp/server-configs-apache/pull/53
Thanks,
Andrew

arthurvr:
@andrewmkhoury Why not to handle it in one PR?

andrewmkhoury:
Sorry, I am still familiarizing myself with github.  I didn't know I could
make changes to a previous PR.
Thanks,
Andrew

andrewmkhoury:
Ok I made a change to my first PR now instead and deleted the other one.

arthurvr:
@andrewmkhoury Thanks. You learned something today :+1:

arthurvr:
@andrewmkhoury Seems like tests are failing, could you take a look?

tx5:
> Sorry, I am still familiarizing myself with github.

> Why not to handle it in one PR?

@arthurvr :-1: give the guy a break, he told you that he's not familiarized with github

AD7six:
Those [test fails](https://travis-ci.org/h5bp/server-configs-apache/jobs/46220699#L1217) are relevant:

> AssertionError: "[an error occurred while processing this directive]\n[an error occurred while processing this directive]\n" == "var test1 = \"test\"\nvar test2 = \"test\"\nvar test3 = \"test\"\nvar test4 = \"test\"\nvar test5 = \"test\"\n\nvar test6 = \"t

I.e. the response was:

    [an error occurred while processing this directive]
    [an error occurred while processing this directive]

instead of the expected:

```
var test1 = "test"
var test2 = "test"
var test3 = "test"
var test4 = "test"
var test5 = "test"

var test6 = "test"
var test7 = "test"
var test8 = "test"
var test9 = "test"
var test10 = "test"
```

According to [the docs](http://httpd.apache.org/docs/2.2/mod/mod_include.html):

> If the directory containing the parsed file has Options IncludesNOEXEC set, then only documents with a text MIME-type (text/plain, text/html etc.) will be included.

So it won't include js files since they are `application/*`. 

Thanks for the suggestion and effort but it looks like I'll have to pass =). If upon further investigation it's possible to make tests pass (i.e. actually achieve the objective) please amend this PR and it can be reopened.

Issue #54, opened 2015-01-15
How about remove X-Powered-By


sheo13666:
```ApacheConf
<IfModule mod_headers.c>
     Header unset X-Powered-By
</IfModule>
```

mathiasbynens:
FTR, it’s better to do this through `php.ini`, by setting `expose_php = off`. Or do other server-side languages than PHP add this header?

sheo13666:
Many project on php add this header in php script, also many hosting company not give set expose_php = off.

mathiasbynens:
@sheo13666 Don’t get me wrong, I agree we could add this to the settings. I’m just saying we should also add a comment explaining the better alternatives.

arthurvr:
> Don’t get me wrong, I agree we could add this to the settings. I’m just saying we should also add a comment explaining the better alternatives.

:+1:

sheo13666:
@arthurvr yes, some later

alrra:
>  Or do other server-side languages than PHP add this header?

@mathiasbynens PHP isn't even the most popular. Using the data from [HTTP archive](http://httparchive.org/): 

| Number of Requests | Value of the `X-Powered-By` Header |
|:---|:---|
| 3858515 | ASP.NET |
| 722266 | PleskLin |
| 120204 | W3 Total Cache/0.9.4.1 |
| 93821 | PHP/5.3.3 |
| 88617 | PHP/5.2.17 |
| 85279 | WAF/2.0 |
| 79098 | PHP/5.3.29 |
| 55817 | W3 Total Cache/0.9.4 |
| 45309 | PHP/5.3.28 |
| 40714 | Express |
| 36329 | Undertow/1 |
| 33041 | Servlet/2.5 JSP/2.1 |
| 29594 | PHP/5.4.36 |
| 27502 | PHP/5.4.35 |
| 25852 | UrlRewriter.NET 2.0.0 ASP.NET |
| 25131 | PHP/5.5.9-1ubuntu4.5 |
| 23518 | ASP.NET ARR/2.5 ASP.NET |
| 22910 | ASP.NET ASP.NET |
| 22361 | PHP/5.3.10-1ubuntu3.15 |


Query used:

  ```sql
SELECT count(*) AS number_of_requests, resp_x_powered_by AS x_powered_by_header_value
FROM [httparchive:runs.2015_02_01_requests]
WHERE resp_x_powered_by != ""
GROUP BY x_powered_by_header_value
[...]

PR #55, opened 2015-01-16
Update server_software_information.conf

sheo13666:


arthurvr:
Would fix https://github.com/h5bp/server-configs-apache/issues/54.

Thanks for the PR, @sheo13666! We should probably add some comments on why we add this. See the other inline comments as examples. Also, from the original issue:

> we should also add a comment explaining the better alternatives.

arthurvr:
friendly ping for @sheo13666

sheo13666:
sorry my English is bad for detail comments :disappointed:

Philosoft:
@sheo13666  напиши по-русски тогда, и я переведу в меру сил)

alrra:
> sorry my English is bad for detail comments :disappointed:

@sheo13666 I took care of this addition, as it required more then a few comments. Thanks again for bringing this up!

PR #56, opened 2015-01-17
Update requests.conf

sheo13666:
set Access-Control-Allow-Origin from host domain

### DIFF
--- a/src/cross-origin/requests.conf
+++ b/src/cross-origin/requests.conf
@@ -9,5 +9,8 @@
 # http://www.w3.org/TR/cors/
 
 <IfModule mod_headers.c>
+    # Only for you domain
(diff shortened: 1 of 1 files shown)


src/cross-origin/requests.conf
```diff
@@ -9,5 +9,8 @@
 # http://www.w3.org/TR/cors/
 
 <IfModule mod_headers.c>
+    # Only for you domain
+    # SetEnvIf Host ^(.*)$ YOU_HOST=$1
```
AD7six: What's the reason for the SetEnvIf call? 

There's also a typo on line 14 `http://%{YOU_HOST}e` <-

alrra:
@sheo13666 Thanks for the pull request, but this example is not useful as you already have access to your own domain. :)

PR #61, opened 2015-02-24
Add WebP to expires rules

philwareham:
There doesn't seem to be a specific WebP image expires rule at present. Adding in.


alrra:
Thanks @philwareham!

Issue #62, opened 2015-03-03
"AddType text/html php" within mod_mime results in blank pages


bridgeport:
First, thank you for maintaining this project. It's been greatly helpful.

Using CentOS 6, Apache 2.4.12, PHP 5.4.38, the following .htaccess line brings down a PHP-powered site -- results in blank pages:
```
AddType   text/html   php
```

Of what I've read about the above, it appears to be proper, but still wanted to report what occurred when I implemented it on my live, vanilla setup.

Thanks.

alrra:
> brings down a PHP-powered site -- results in blank pages

@bridgeport Does adding:

```apache
<FilesMatch \.php$>
    SetHandler application/x-httpd-php
</FilesMatch>
```

fix the issue by any chance?

-

> Of what I've read about the above, it appears to be proper

Yes, the `text/html` media type is used even by sites like [php.net](https://php.net/).

e.g.:

```bash
$ curl -sSIL https://php.net/downloads.php

HTTP/1.1 200 OK
Server: nginx/1.6.2
Date: Tue, 03 Mar 2015 09:18:11 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive
X-Powered-By: PHP/5.6.5-1
Content-language: en
X-Frame-Options: SAMEORIGIN
Set-Cookie: COUNTRY=NA%2C86.125.178.69; expires=Tue, 10-Mar-2015 09:18:11 GMT; Max-Age=604800; path=/; domain=.php.net
Cache-Control: private
Pragma: no-cache
Link: <http://php.net/downloads>; rel=shorturl
```

alrra:
@bridgeport Thanks for bringing this up. It seems that there is a need to add more PHP specific¹  configs, and that is beyond the scope of this project, so the `AddType text/html php` line is now [gone](https://github.com/h5bp/server-configs-apache/commit/daab35bc91c299da23efdfa740f13a46ad10612a). 

--

¹  https://php.net/manual/ro/install.unix.apache2.php
  http://blog.famillecollet.com/post/2013/01/13/PHP-and-Apache-SetHandler-vs-AddHandler

PR #63, opened 2015-03-03
Correcting .htaccess file

bpoinsot:
"AddType text/html php" isn't a good idea IMO ^^


arthurvr:
Why wouldn't that be a good idea? Please try to provide some additional information.

arthurvr:
Fixed in https://github.com/h5bp/server-configs-apache/commit/daab35bc91c299da23efdfa740f13a46ad10612a.

alrra:
@bpoinsot Thanks, but I already [took care of this](daab35bc91c299da23efdfa740f13a46ad10612a). 


@bpoinsot ,@arthurvr See also: h5bp/server-configs-apache#62.

PR #64, opened 2015-04-07
Set X-UA-Compatible only in IE

szepeviktor:


alrra:
@szepeviktor See https://github.com/h5bp/server-configs-apache/pull/26#issuecomment-41156774.

szepeviktor:
Thank you.

PR #69, opened 2015-06-01
Update .htaccess

mattbrundage:
rewrote part of the comment for "X-Powered-By" response header


AD7six:
Looks fine - you'll need to change [the source file](https://github.com/h5bp/server-configs-apache/blob/master/src/security/x-powered-by.conf) - otherwise, that change will be overwritten the next time there's a release.

mattbrundage:
Done.

arthurvr:
You forgot to edit another file as well. Will update it while merging.

arthurvr:
Thanks!

PR #74, opened 2015-06-27
gitattributes: override *.conf using linguist-language=ApacheConf

hatzopoulos:
Helps to recognize .conf files as ApacheConf files in GitHub as
described at https://github.com/github/linguist/blob/master/README.md

Currently this repo has 43% JavaScript and 40% Apache files.


alrra:
@hatzopoulos Thanks! :purple_heart:

PR #79, opened 2015-09-29
Block Mercurial .orig files

walterebert:
Mercurial creates a backup file with the extension .orig when a file is reverted. This could lead to a potential leak of sensitive information.


XhmikosR:
LGTM but I'm not sure if `.orig` is being used for other stuff too. Do we have any more info on this?

walterebert:
It seems that `.orig` can also be used in git and diff tools (https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files). These serve the same purpose as in Mercurial.

PR #83, opened 2015-10-29
edit "Strict-Transport-Security" and use "upgrade-insecure-requests"

voku:
[+](edit): Upgrading legacy sites to HTTPS, via "upgrade-insecure-requests"

[Review on Reviewable](https://reviewable.io/reviews/h5bp/server-configs-apache/83)


LeoColomb:
This change is redundant with the previous part for CSP: everything needs to be configured, even for `upgrade-insecure-requests` tag.

https://github.com/h5bp/server-configs-apache/blob/4c1364885477e836fec24a6d8330cba69cf3d3a0/src/security/content-security-policy.conf#L11-L16

Maybe an elaboration can be added in the comment to explain, but as there's already a lot of good documentation about CSP I'm not sure it's really useful.

About HTST, I wonder if having two different HSTS directives will not a bit confusing for users...

Though @XhmikosR?

XhmikosR:
Yup, I agree with @LeoColomb

PR #84, opened 2015-11-03
fix work forcing www/no-www with https

### LINKED ISSUE #52: Remove only-http from www-normalizing redirects
Most users probably aren't using both http and https, but if they are it's possibly confusing/misleading why these redirects only apply to http requests.

So, as mentioned in #49 the `RewriteCond %{HTTPS} =on` conditions should be removed before the next major release.

xtratio:
Problem: Options "Suppressing / Forcing the `www.` at the beginning of URLs" only work with http.
Hope this will be helpful.

Fixes #52

### DIFF
--- a/src/rewrites/rewrite_www.conf
+++ b/src/rewrites/rewrite_www.conf
@@ -22,7 +22,6 @@
 
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteCond %{HTTPS} !=on
     RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]
     RewriteRule ^ %{ENV:PROTO}://%1%{REQUEST_URI} [R=301,L]
 </IfModule>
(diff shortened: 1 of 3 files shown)


lucaspetter:
Agreed. This should be merged. :+1:

Rewrites that force/suppress the `www` should be independent of rewrites that force/suppress `https`. It should be possible to force/suppress `www` regardless of whether the protocol is `https` or `http`. This commit fixes the current dependence on `http` for forcing/suppressing `www`.

I came here to make the exact same pull request, but am glad to see someone else has already done it.

XhmikosR:
Discussion on this: https://github.com/h5bp/server-configs-apache/issues/52

ManasN:
This should be merged along with a comment, really. I knew they have some intention to put it there but couldn't figure out why, now that #52 makes it clear, it will be helpful if we keep this in the .htaccess.

src/rewrites/rewrite_www.conf
```diff
@@ -22,7 +22,6 @@
 
 <IfModule mod_rewrite.c>
     RewriteEngine On
-    RewriteCond %{HTTPS} !=on
```
LeoColomb: @xtratio Could you add a comment in the heading part as @ManasN (https://github.com/h5bp/server-configs-apache/pull/84#issuecomment-289069219)?

LeoColomb:
Closed #84 via fc747bbdf0a0c224ec08d8b925f33671e4d5046d.

lucaspetter:
Thank you, @LeoColomb !

PR #87, opened 2015-12-06
only set X-UA-Compatible header if html document

Computer-Whiz123:
modified lines 132-133 to only set the X-UA-Compatible header if it is
a html file as per the comments instead of setting it and un-setting it
for non-HTML files

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -124,14 +124,12 @@ Options -MultiViews
 
 <IfModule mod_headers.c>
 
-    Header set X-UA-Compatible "IE=edge"
-
     # `mod_headers` cannot match based on the content-type, however,
     # the `X-UA-Compatible` response header should be sent only for
     # HTML documents and not for the other resources.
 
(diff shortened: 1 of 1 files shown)


QWp6t:
Nah, this won't work. Header won't be set for anything that isn't explicitly `.html`. So that means that `.php` would be excluded, for example. Even `.htm` would be excluded.

Seting it by default and conditionally removing it makes a lot more sense since it's much more flexible.

alrra:
@Computer-Whiz123 Thanks for the pull request, but as @QWp6t correctly pointed out, there are a lot of possible extensions for HTML documents (to name a few more:  `.aspx`, `.jsp`, `.do`, `.cgi`) 

See also: https://github.com/h5bp/server-configs/issues/113#issuecomment-12370812.

Computer-Whiz123:
good to know. This was just a quick-fix but i'll update it with other extensions...

alrra:
>  i'll update it with other extensions...

@Computer-Whiz123 No need to do that. As I said [before](https://github.com/h5bp/server-configs-apache/pull/87#issuecomment-162286041), this is [not a good solution](https://github.com/h5bp/server-configs/issues/113#issuecomment-12370812) as maintaining a list of all possible extensions will be difficult, plus, we'll end up with an even bigger list than the currently used one.

> I'm still kinda new at this whole github thing...

@Computer-Whiz123 Don't worry! :)

PR #88, opened 2015-12-06
added other filetypes to set the header for

Computer-Whiz123:
see: https://github.com/h5bp/server-configs-apache/pull/87


alrra:
Closing because of the reasons stated in https://github.com/h5bp/server-configs-apache/pull/87#issuecomment-162286041.

PR #109, opened 2017-04-25
Prevent forcing https on local hosts

giovannipds:
Hey.
I don't know what you guys think but, I think it'll never be necessary to have local hosts forcing https, so I've included the same lines contained on `src/rewrites/rewrite_www.conf` to the `src/rewrites/rewrite_http_to_https.conf` file.
I couldn't run the tests here, mainly probably because of the outdated devDependencies, so I've edited manually on the dist file as well.
If you do not agree by any reason, I'd like to explain me why, if possible.
Thank you guys. =)

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -352,6 +352,8 @@ AddDefaultCharset utf-8
 # <IfModule mod_rewrite.c>
 #    RewriteEngine On
 #    RewriteCond %{HTTPS} !=on
+#    RewriteCond %{SERVER_ADDR} !=127.0.0.1
+#    RewriteCond %{SERVER_ADDR} !=::1
 #    RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
 # </IfModule>
 
--- a/src/rewrites/rewrite_http_to_https.conf
+++ b/src/rewrites/rewrite_http_to_https.conf
@@ -8,5 +8,7 @@
 <IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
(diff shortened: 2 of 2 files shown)


XhmikosR:
@LeoColomb: thoughts?

I personally don't think this belongs here since it's for development only.

LeoColomb:
> I personally don't think this belongs here since it's for development only.

That's exactly the words I was looking for, thanks! 😆 
So, yes, agree.

giovannipds:
I'm sorry guys, but I may be not understanding your thoughts. Why denying this? I know it's for development only, but, If I'd like to keep the same `.htaccess` between my environments (not that this is a good practice) the rule without the condition lines I've proposed will force HTTPS even on local development (which I guess it'll never be the case, will it?). If it'll never exist HTTPS on local, it'd be better to have that on the condition... wouldn't it? Am I missing something here? cc @XhmikosR @LeoColomb cheers!

LeoColomb:
> not that this is a good practice

😇

My complete thought is that if you're on dev environment using Apache either you simply disable https redirection, either you create a local certificate. But I may be wrong...
Actually, I'm also wondering about all proxy-like software submitting requests to apache on localhost (Varnish?), should we consider the proxy handles https redirection?

giovannipds:
@LeoColomb I just don't understand why you guys are using a different discernment when [rewrite_www.conf](https://github.com/h5bp/server-configs-apache/blob/master/src/rewrites/rewrite_www.conf) uses the same logic. It's helpful, it's logic and won't bother. 2 lines. But I won't be begging for this. You don't to, that's ok.

I don't know about the proxy-like softwares you've mentioned.

PR #110, opened 2017-04-25
Change the rule of forcing https to keep up the request uri

giovannipds:
Hi again.
I had a case where something like `http://www.domain.com/subdirectory/` was redirecting to `https://www.domain.com`, by the actual project code for https force, located on `src/rewrites/rewrite_http_to_https.conf`:

```apache
RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
```

While testing, I made this update which has fixed the issue:

```apache
RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
```

=) anything, let me know. Cheers.

Note: I couldn't run the tests here (devDependencies outdated).

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -352,7 +352,7 @@ AddDefaultCharset utf-8
 # <IfModule mod_rewrite.c>
 #    RewriteEngine On
 #    RewriteCond %{HTTPS} !=on
-#    RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
+#    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 # </IfModule>
 
 # ----------------------------------------------------------------------
--- a/src/rewrites/rewrite_http_to_https.conf
+++ b/src/rewrites/rewrite_http_to_https.conf
@@ -8,5 +8,5 @@
 <IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
-   RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
+   RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 </IfModule>


LeoColomb:
Actually, it doesn't change some much the redirection behavior, but for consistency with HTTP to HTTPS redirection, I would merge this.
That said I wonder if the official example (https://wiki.apache.org/httpd/RewriteHTTPToHTTPS) is better or not.

giovannipds:
@LeoColomb I've never tried using `%{SERVER_NAME}` but learned `%{REQUEST_URI}` from the www redirects [here](https://github.com/h5bp/server-configs-apache/blob/master/src/rewrites/rewrite_www.conf). If they're working there, I thought they would work well here too.

My knowledge on apache configs are very limited, but if we were using `(.*)` until now, they weren't working as expected on the case I've mentioned (and the official example is very similar, despite using the leading slash, the optional `?` character and `%{SERVER_NAME}`). I couldn't find anything easy to read about `%{REQUEST_URI}` too, but I'd use this until anything different's mentioned. Thanks for approving and review this.

XhmikosR:
I'm confused too. You mean the official docs are wrong since they are using `RewriteRule ^/?(.*) https://%{SERVER_NAME}/$1 [R,L]`?

Using `REQUEST_URI` seems to make sense here indeed. I wonder if there are any extra redirects? Did anyone test this yet?

giovannipds:
If they're "wrong" or not, I can't guarantee, that's not to me to judge. Something they do, that's sure, but what I tested, did it partially. I'm also not sure how much these three are different from each other:

```apache
RewriteRule ^(.*)$ https://%{HTTP_HOST}/$1 [R=301,L]
# which didn't work as expected when I tested
# in a production server (as described above)
    
RewriteRule ^/?(.*) https://%{SERVER_NAME}/$1 [R,L]
# from the official docs

RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
# with REQUEST_URI, which has worked to me
# I still think this is the most generic way possible, not
# forcing leading slash usage (which cause different
# behavior with directories, with R=301, which is correct, etc.)
```

Maybe we can think some url cases to test this out, I don't know...

PS.: I thought this project was abandoned.

LeoColomb:
I've done some tests, and I conclude as @giovannipds. Merging.
Thanks @giovannipds for your good explanations! 👍 

>  I wonder if there are any extra redirects?

Not for me, just perfect redirections.

> Maybe we can think some url cases to test this out, I don't know...

Yep, maybe with #45.

giovannipds:
@LeoColomb nice to know that, I'm glad to help. Thanks for your attention and availability.

PR #114, opened 2017-09-16
Font MIME types following RFC 8081

LeoColomb:
Recent changes introduced by the [RFC 8081](https://tools.ietf.org/html/rfc8081) and [registered at IANA](https://www.iana.org/assignments/media-types/media-types.xml#font).

Current|New
---|---
`application/x-font-ttf`|`font/ttf`
`font/opentype`|`font/otf`
`application/font-woff`|`font/woff`
`application/font-woff2`|`font/woff2`
`application/x-font-ttf` (ttc)|`font/collection`

### DIFF
--- a/src/media_types/media_types.conf
+++ b/src/media_types/media_types.conf
@@ -55,20 +55,11 @@
 
   # Web fonts
 
-    AddType application/font-woff                       woff
-    AddType application/font-woff2                      woff2
+    AddType font/woff                                   woff
+    AddType font/woff2                                  woff2
     AddType application/vnd.ms-fontobject               eot
-
-    # Browsers usually ignore the font media types and simply sniff
-    # the bytes to figure out the font type.
-    # https://mimesniff.spec.whatwg.org/#matching-a-font-type-pattern
-    #
-    # However, Blink and WebKit based browsers will show a warning
-    # in the console if the following font types are served with any
-    # other media types.
-
-    AddType application/x-font-ttf                      ttc ttf
-    AddType font/opentype                               otf
+    AddType font/ttf                                    ttc ttf
+    AddType font/otf                                    otf
 
 
   # Other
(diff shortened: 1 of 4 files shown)


mathiasbynens:
Which browsers support these MIME types?

LeoColomb:
At least the latest version of modern browsers, but this is hard to say except not being able to produce an error.
That said, browsers [usually](https://mimesniff.spec.whatwg.org/#matching-a-font-type-pattern) ignore the font media types and simply sniff the bytes to figure out the font type.

Also, as an example, Google Fonts already use these "new" types:

```
$ curl -I https://fonts.gstatic.com/s/lato/v13/9k-RPmcnxYEPm8CNFsH2gg.woff

HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Type: font/woff
[...]
```
_Extracted from [Lato](https://fonts.googleapis.com/css?family=Lato) served for IE_

```
$ curl -I https://fonts.gstatic.com/s/lato/v13/UyBMtLsHKBKXelqf4x7VRQ.woff2

HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Type: font/woff2
[...]
```
_Extracted from [Lato](https://fonts.googleapis.com/css?family=Lato) served for Chrome_

mathiasbynens:
https://github.com/jshttp/mime-db/issues/77

LeoColomb:
This is a very good link, thanks!

But this seems to be kind a vicious circle: if everyone wait to the other be updated, no one will do.
Most of the time it can be understandable to avoid errors and breaking things, but in our case the worst thing this change could raise is a notice in some old browsers. 🤔 

Isn't it? What is your opinion?
🙂

mathiasbynens:
Yeah, I agree this should be merged! Could you please take a look at the CI failure?

mathiasbynens:
Any reason why `font/sfnt` is not added?

LeoColomb:
Rebased.

mathiasbynens:
👍

XhmikosR:
@LeoColomb: Shouldn't we remove the old types completely?

LeoColomb:
I would keep them for compatibility as software may force them. 
But that said, I don't know at all.

PR #120, opened 2017-09-30
Add template for files match pattern

### LINKED ISSUE #104: Pattern placeholders
Could we set up some kind of templating?

Some matching patterns appear more than once.
https://github.com/h5bp/server-configs-apache/blob/58d58408821f3ade93cae09cd33fdfa4b376d533/src/security/x-xss-protection.conf#L46

LeoColomb:
Fixes #104 (=> DRY)
Makes it easier to follow changes in this pattern.

### DIFF
--- a/bin/build/build.sh
+++ b/bin/build/build.sh
@@ -102,6 +102,8 @@ create_htaccess() {
     insert_line "" "$file"
     insert_file_comment_out "src/web_performance/filename-based_cache_busting.conf" "$file"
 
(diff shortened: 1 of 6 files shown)


XhmikosR:
Should this be in src and not in bin/build?

LeoColomb:
I think yes since this is not part of build scripts but part of the "active" content.

szepeviktor:
Thank you @LeoColomb

PR #128, opened 2017-11-17
[FEATURE] Add htaccess directives suggestions for HSTS preload

afoeder:
This adds some suggestions for enabling HSTS preload support
via adjusting the appropriate headers sent via the .htaccess
apache configuration.

### DIFF
--- a/src/security/strict-transport-security.conf
+++ b/src/security/strict-transport-security.conf
@@ -13,13 +13,35 @@
 # server via HTTPS, regardless of what the users type in the browser's
 # address bar.
 #
+# (!) Be aware that this, once published, is not revokable and you must ensure
+# being able to serve the site via SSL for the duration you've specified
+# in max-age. When you don't have a valid SSL connection (anymore) your
+# visitors will see a nasty error message even when attempting to connect
+# via simple HTTP.
+#
 # (!) Remove the `includeSubDomains` optional directive if the website's
 # subdomains are not using HTTPS.
 #
+# (1) If you want to submit your site for HSTS preload (2) you must
+#     * ensure the `includeSubDomains` directive to be present
+#     * the `preload` directive to be specified
+#     * the `max-age` to be at least 31536000 seconds (1 year) according to the current status.
+#
+#     It is also advised (3) to only serve the HSTS header via a secure connection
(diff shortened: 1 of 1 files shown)


src/security/strict-transport-security.conf
```diff
@@ -16,10 +16,27 @@
...
+#
+#     It is also advised (3) to only serve the HSTS header via a secure connection
+#     which can be done with either `env=https` or `"expr=%{HTTPS} == 'on'"` (4). The
+#     exact way depends on your environment and might just be tried.
+#
 # https://www.html5rocks.com/en/tutorials/security/transport-layer-security/
 # https://tools.ietf.org/html/draft-ietf-websec-strict-transport-sec-14#section-6.1
 # https://blogs.msdn.microsoft.com/ieinternals/2014/08/18/strict-transport-security/
+# (2) https://hstspreload.org/
+# (3) https://tools.ietf.org/html/draft-ietf-websec-strict-transport-sec-14#section-7.2
```
LeoColomb: Is it possible to use the final RFC for the link? https://tools.ietf.org/html/rfc6797
In the meantime, could be good to update also the other rfc6797 just above.

src/security/strict-transport-security.conf
```diff
@@ -16,10 +16,27 @@
...
+# (4) https://stackoverflow.com/questions/24144552/how-to-set-hsts-header-from-htaccess-only-on-https/24145033#comment81632711_24145033
+#
```
LeoColomb: Just a visual change here: would be better without this `#` empty line. 😉 💯

XhmikosR:
My main concern  is that we shouldn't hardcode the requirements like that.
I mean previously they were different, now they are like that etc. So in my
opinion we should just emphasize more that the user should checkthe HSTS
preload site at a given time for the exact requirements.

src/security/strict-transport-security.conf
afoeder: good points, will do!

afoeder:
@XhmikosR, I see your point. Since I'd say we're still rather "fluent" with upstream changes especially in Open Source _I_ would suggest leaving the out-commented code line as-is and emphasize/retext the remarks accordingly to note to better check, or sthg like this.
WDYT?

afoeder:
I've adjusted the code, please re-review.

Issue #89, opened 2015-12-11
Proposal: Customizable builds?


tomkyle:
The build.sh script concats a number of ".conf" modules, some of which are commented out. So each time a new version is released, developers will copy the dist/.htaccess to their project root and will have to scroll through and seek the modules they last time "commented in" again (or vice versa), e.g. the HSTS section. Would it not be fine if we could pass a "build configuration" to the build script?

A simple config file could roughly look like this, e.g. as an bash include:

``` bash
# build.conf.inc

declare -A modules
# true:  An active module
# false: module to be commented out
modules["src/cross-origin/requests.conf"]=false
modules["src/cross-origin/images.conf"]=true
modules["src/cross-origin/resource_timing.conf"]=true
modules["src/cross-origin/web_fonts.conf"]=false
# ...
```

The `build.sh` script would then use a tweaked insert function, for example:

``` bash
# Import configuration in "modules" var, i.e.
# fill OPTIONS with CLI param $1 
# or default to build.conf.inc 
source $OPTIONS

# Includes or "includes-and-comments-out":
insert_file_if() {
    if [ ${modules["$1"]} = true ]; then
        insert_file "$1" "$2"
    else
        insert_file_comment_out "$1" "$2"
    fi
}

# Uses new insert_file_if function:
create_htaccess() {
    insert_header "cross-origin" "$file"
    insert_line "" "$file"
    insert_file_if "src/cross-origin/requests.conf" "$file"
    insert_line "" "$file"
    insert_file_if "src/cross-origin/images.conf" "$file"
    # ...
}
```

What do you think?

LeoColomb:
Very interesting! Good idea!
Could you make a PR? 😃

tomkyle:
I'll do my very best :-D (I'm afraid though, I'll have to ask my time machine, after 20 months and several HD tidyups)

tomkyle:
Sorry folks, I have not forgotten you. ...putting this back on stack.

LeoColomb:
@tomkyle No problem, take your time (even if I can't wait to see your amazing PR!) 😺 👍

LeoColomb:
Thanks a lot for your impressive contribution @tomkyle, very appreciated! 🎉 👍

tomkyle:
It was a pleasure :-) Thank you @LeoColomb for the advice here and there!

PR #141, opened 2018-03-14
Customizable builds: a first approach

tomkyle:
This PR addresses #89 “Customizable builds”:

- Added a ,partials‘ configuration file _partials.conf_ which defines enabled or disabled (commented-out) htaccess module parts.
- Changed the _bin/build/build.sh_ script to utilize this file.
- The associative array introduced in _partials.conf_ requires Bash 4. So a `#!/usr/bin/env bash` shebang becomes necessary in _build.sh_


tomkyle:
Well, checks have failed... not quire sure what Travis wants to tell me. Can someone hint me in the right direction?

LeoColomb:
This seems a really really good start!
Amazing, thanks @tomkyle! 🎉 

Regarding build fail, this is because "fixture" variant used for tests is different from the default one.
Actually, could be really interesting to use a config file for the "fixture" variant.

Also, a quick note before full review: 
* May be prefer a loop over the `partials` array instead of keeping all file references in `bin/build/build.sh` (would allow to add custom source files 😍)

tomkyle:
Thx, Leo, for your explanation. Of course, fixtures build is different than the one for _dist/_, I've completely overlooked this. I'll try to fix this. 

Noob question (I'm new in this business): When I commit to my fork branch, the PR updates automatically, doesn't it?

LeoColomb:
> When I commit to my fork branch, the PR updates automatically, doesn't it?

Yes.

tomkyle:
We now have two nearly idendical functions,  _create_htaccess_ and _create_htaccess_fixture_. Their main differences are:

- The intro text “Using \`.htaccess\` files slows down Apache…” is only used for dist/htaccess, 
- Partials used in _create_htaccess_fixture_ are always active. 

And we now have two build config files, _fixtures.conf_ and _partials.conf._ I feel this should undergo a refactoring. I see two directions I'd like to discuss:

1. Refactor both functions into one _create_htaccess_ function. The intro text therefore would go into a new partial file, _src/intro.conf_, which then would be referenced from the build config file. Still two config files, each of them to be kept up-to-date.

2. If the fixtures build is actually meant to provide each partial _always enabled,_ we could save the fixtures config. The fixtures build function would then simply parse the partials.conf and always assume “enable” for the action keyword. So we had only one config file left which I feel is more future-proof; we of course still woudl have had a _create_htaccess_fixture_ on its own like before but with a slightly shorter parsing section.  

As my PR is awaiting review, I'm not sure how to proceed (that's the Git beginner in me): Should I await your comments first, @LeoColomb, or would you mind if I go on writing and committing?

LeoColomb:
I definitely prefer the generic option 1., with only one `create_htaccess()` function:

> The intro text therefore would go into a new partial file, src/intro.conf, which then would be referenced from the build config file.

No, we basically don't care that much of the intro text, it could be added at each build even for the fixture one. We must avoid too much settings complexity.

> The fixtures build function would then simply parse the partials.conf and always assume “enable” for the action keyword.

This can change in time. Also, for future test we may use this new build system to test different things separately, with different small build config files (or may be not, but let say).

---

> Should I await your comments first, @LeoColomb, or would you mind if I go on writing and committing?

While the PR is not merged, feel free to add commit at any here. Review may be blocking for maintainers but never for the contributor. ([see doc](https://guides.github.com/introduction/flow/) 😃)

LeoColomb:
Amazing, everything looks good! 🎉 

I will wait one or two days before merging to let @XhmikosR review this PR if needed, and to let you @tomkyle adding<sup>1</sup> a bit of a documentation about this new feature on the README and if you feel courageous adding a really simple test over this change (basically testing each config keyword).

Thanks! 😺 

---

<small><sup>1</sup> If you're not really fluent with Git as you said before, this is just a small reminder that it may be better if you `git pull` before working again on your local branch. Happy coding! ⌨️ </small>

tomkyle:
Thank you for your additions; it is an excellent idea to leverage the main function and to split the build calls. Proposal: what if we changed the parameter order for the build script? This would allow custom output locations from a default config:

```
$ .../build.sh path/to/project/.htaccess
# or with project config:
$ .../build.sh htdocs/.htacess ./htaccess.conf
```

This leads me to the idea to set `htaccess_output_default` to `./.htaccess` and to seek for `htaccess_config_default` in `$PWD` first. What do you think?

LeoColomb:
@tomkyle As you want. That said your argument seems relevant. 🙂

bin/build/build.sh
```diff
@@ -256,20 +158,26 @@ print_success() {
 # ----------------------------------------------------------------------
 
 main() {
+    local htaccess_output="${1}"
+    local htaccess_config="${2}"
+    local htaccess_output_directory="$(dirname "${htaccess_output}")"
 
-    cd "${BASH_SOURCE%/*}" && cd ../../
+    if [ ! -f "${PWD}/${htaccess_config_default}" ]; then
```
LeoColomb: -> `"${PWD}/${htaccess_config}"`

tomkyle: Thank you, Sir!

tomkyle:
I'm not quite sure if my documentation draft/“custom builds” section is too lengthy, or even understandable enough. Any feedback on this will be appreciated. 

Since I'm no native English speaker, I invite everyone to also have a thorough look for uncommon sentences or unorthodox phrases…

For now, it's supper time.

bin/build/build.sh
```diff
@@ -3,6 +3,31 @@
...
+    if [ ! -d "${temp_directory}" ]; then
+        print_error "Creating temp directory failed."
+        exit 1
+    fi
+}
+
+# Default Exit or SIGINT(2) handler
+trapCleanupTempDir() {
+    [ -d "${temp_directory}" ] \
+    && rm -Rf "${temp_directory}" \
```
LeoColomb: I don't like at all your last commit...

This commit adds too much extra complexity, and this command is really much more risky than having a failed build.

Let user handle their filesystem and temp build files. 😅

bin/build/build.sh
```diff
@@ -1,19 +1,51 @@
...
 # ----------------------------------------------------------------------
 # | Helper functions                                                   |
 # ----------------------------------------------------------------------
 
 clean() {
-    rm -rf dist/
-    rm -rf test/fixtures/.htaccess
+    for removee in "$@"
+    do
+        rm -rf "${removee}"
```
LeoColomb: Actually recursive flag can be removed here, I think. It would avoid dumb user to delete the whole system! 🤣

README.md
```diff
@@ -56,6 +58,111 @@ use them, please check the appropriate Apache documentation:
 * <https://httpd.apache.org/docs/current/configuring.html>
 * <https://httpd.apache.org/docs/current/howto/htaccess.html>
 
+## Custom .htaccess builds
+
+Security, mime-type, and caching best practices evolve, and so should do your *.htaccess* file. In the past, with each new *Apache Server Configs* release it was quite tedious to find out which *.htaccess* trick was just new or only had changes in certain nuances.
+
+Our new [**build script**](#build-script-buildsh) with its re-usable and customizable [**build configuration**](#configuration-file-htaccessconf) lets you easily update your *.htaccess* file. Each new *.htaccess* build will contain our updated source files, enabled or commented-out according to your settings in the *htaccess.conf* of your project root.
```
LeoColomb: Well, the build script is new right now, yup, but will not be in the future. 😃 

Also, sentence without attribution are often better (Our -> The).

README.md
```diff
@@ -56,6 +58,111 @@ use them, please check the appropriate Apache documentation:
...
+
+### Configuration file: *htaccess.conf*
```
LeoColomb: Good part! 👍

bin/build/build.sh
```diff
@@ -206,6 +124,15 @@ insert_file_comment_out() {
     printf "%s\n" "$(cat "$1" | sed -E 's/^([^#])(.+)$/# \1\2/g')" >> "$2"
 }
 
+# Includes or "includes-and-comments-out":
+insert_file_if() {
```
LeoColomb: This function is no more used, right? 🙂

bin/build/build.sh
```diff
@@ -223,11 +150,11 @@ insert_space() {
 
 apply_pattern() {
     sed -e "s/%FilesMatchPattern%/$( \
-        cat "src/files_match_pattern" | \
+        cat "${repo_root}/src/files_match_pattern" | \
         sed '/^#/d' | \
         tr -s '[:space:]' '|' | \
         sed 's/|$//' \
-    )/g" --in-place "$1"
+    )/g" -i "" "$1"
```
LeoColomb: 👍

README.md
tomkyle: Valid argumentation :-) Agree. Applied.

bin/build/build.sh
tomkyle: Yep. Will be removed.

bin/build/build.sh
tomkyle: At first, thank you for your review. It must be hard work to look through dozens od commits... 

Please allow me to plead: Project roots, especially on live environments, are not the place where to risk cluttered file system. See, it is widely-used practice to not checkout .htaccess with the repo, and to maintain the .htaccess on each enviroment individually (for example: rewrite_base directives). So developers will likely checkout their conf file instead, and build their production .htaccess with our build tool. If only one small thing fails, the web app can be considered broken for at least minutes. This is why I'd prefer a “sandbox” for building. 

You indeed are right, the trap EXIT SIGINT stuff is complex (and surely not seen very often). But e.g. on shared hosting environments I'd prefer to not have temp builds lying around. This was my reason for “use temp, but tidy up after all”. 

If we omitted this functionality, well, we would have to: 

- Check for existing .htaccess file first
- Create backup
- Try build process
- If it fails, restore .htaccess from backup.

This steps alone relocate the “too-much complexity” back into main(), so no use here.

And: currently, the create function 'exits 1' without discussion when something fails. In contradiction, backup restauration implies returning back to main function, adding even more complexity in return code evaluation.

Of course, users should have their filesystem on radar. But all this boilerplate stuff can be tempting to not have. Some people might even confuse it with “not need to” ;-)

bin/build/build.sh
tomkyle: Good one. It's not used anyway. To be removed.

README.md
tomkyle: Thx :-)

bin/build/build.sh
tomkyle: :-) When to best close that issue? Right now, or after this branch has been merged?

tomkyle: Oh, closed already. Thx!

bin/build/build.sh
tomkyle: I still see your point, @LeoColomb. What about a reasonable compromise? Commit #56d1eeb decreases output verbosity and removes redundant directory checks.

tomkyle: Finally, I found a way to go without the trap function, and to just work with a tmp file right beneath the final .htaccess. In theory this has the flaw that in case of error exits, after mkdir -p, there might be a newly created path/to/custom/ directory which is in fact empty. 

Anyhow, I ran into some trouble with Travis. Please see new comment below.

tomkyle:
OK, so we go without trap function. Create function now writes to a temp file right beneath the final .htaccess. Buuuut, the Travis build log shows some funny:

```
sed: can't read : No such file or directory
 [✖] Build .htaccess 
 [✔] Moved in place: 'path/to/another/subdir/.htaccess'
```
OK then, sed has some issues. I could not figure out where exactly. 
But it's plausible that $create_htaccess_result is falsy (2), and thus the red cross.
But why (or how?) can the temp output then be moved in place?

I'm a bit perplexed. Could you have a look, @LeoColomb?

bin/build/build.sh
LeoColomb: > This is why I'd prefer a “sandbox” for building.

Sorry for my late reply, but I still think what you call sandbox for build must be managed by the user, not by the building tool.

* More and more deployment pipelines required build step to be separated from the deploy step, even more when using CI/CD.
* This add some system-relative complication, I mean are you sure that the directories used outside of the repo's one are always present?
* This may generate more issues regarding deployment, with users expecting a perfect deployment script, which is not our goal and where I definitely prefer adding a warning notice.

IMHO, a check step before clean/build is more relevant. Checking if all files are available and readable.

tomkyle: Thank you for rour reply. It hasn't been too long since, and btw. my three-day-holidays are over, so my commit frequency will melt down to a average spare time level again ;-)

Regarding our discussion, these are arguments I can follow. (External temp dir usage has already been removed with one *cough* of the latest commits.)

- I like the idea with a before-check if the files (or the target directory) in question can be written at all. “…adding a warning notice” – we could check before `mkdir -p`, if $PWD is writeable.
- Can we safely assume the dir created with `make -p` is writeable? I think so.
- Raising an error when .htaccess exists is not necessary IMHO; Here I would prefer the backup created, together with a notice on this (like it is done now), because creating/updating an htaccess is exactly what this script is supposed to do. Users will know that there may exist one. Perhaps just a matter of taste?

PR #149, opened 2018-04-25
Update x-ua-compatible.conf

mattbrundage:
Set this header only in Internet Explorer (other browsers have no use for it.)

### DIFF
--- a/src/internet_explorer/x-ua-compatible.conf
+++ b/src/internet_explorer/x-ua-compatible.conf
@@ -7,6 +7,9 @@
 #
 # https://hsivonen.fi/doctype/#ie8
 #
+# Override Compatibility View settings in Internet Explorer 11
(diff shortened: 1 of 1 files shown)


alrra:
@mattbrundage See https://github.com/h5bp/server-configs-apache/pull/26#issuecomment-41156774.

mattbrundage:
From #26:
> what if the user is behind a proxy that overrides or mangles the UA string?

@mathiasbynens , are there any known examples of proxies doing this, specifically, removing the string "`Trident`" from IE's UA string? If there are no known examples, I'm left with having to [prove a negative](https://en.wikipedia.org/wiki/Burden_of_proof_(philosophy)).

LeoColomb:
> are there any known examples of proxies doing this

You can't even imagine how many wrongly configured proxies are used, seriously...

As the comment pointed out, it worth having a unused header than a broken website.

PR #157, opened 2018-06-12
Allow SSL certificate set up over HTTP

westhouseit:
A non-HTTPS connection is required in some cases.

### DIFF
--- a/src/rewrites/rewrite_http_to_https.conf
+++ b/src/rewrites/rewrite_http_to_https.conf
@@ -18,7 +18,7 @@
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
    # (1)
-   # RewriteCond %{REQUEST_URI} !^/\.well-known/acme-challenge/[0-9a-zA-Z_-]+$
+   # RewriteCond %{REQUEST_URI} !^/\.well-known/acme-challenge/
    # RewriteCond %{REQUEST_URI} !^/\.well-known/cpanel-dcv/[0-9a-zA-Z_-]+$
    # RewriteCond %{REQUEST_URI} !^/\.well-known/pki-validation/[A-F0-9]{32}\.txt(?:\ Comodo\ DCV)?$
    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
--- a/src/rewrites/rewrite_http_to_https.conf
+++ b/src/rewrites/rewrite_http_to_https.conf
@@ -9,5 +9,8 @@
 <IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
+   RewriteCond %{REQUEST_URI} !^/\.well-known/acme-challenge/[0-9a-zA-Z_-]+$
+   RewriteCond %{REQUEST_URI} !^/\.well-known/cpanel-dcv/[0-9a-zA-Z_-]+$
+   RewriteCond %{REQUEST_URI} !^/\.well-known/pki-validation/[A-F0-9]{32}\.txt(?:\ Comodo\ DCV)?$
    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 </IfModule>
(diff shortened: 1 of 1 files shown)


LeoColomb:
Thanks for opening this PR @westhouseit!
Do you have any documentation regarding these endpoints?

westhouseit:
Basically it's illogical to do file-based/webroot validation over HTTPS when issuing a certificate since this may be the first cert for the server.

But anyway, here's some links clarifying all three lines in the pull request: 
https://letsencrypt.readthedocs.io/en/latest/using.html#webroot
https://support.cloudflare.com/hc/en-us/articles/214820528-How-to-Validate-a-Let-s-Encrypt-Certificate-on-a-Site-Already-Active-on-Cloudflare : "Be sure to check the page rules for the domain in the Cloudflare dashboard and verify that there aren't any which would result in a request to the validation URL to be **redirected or only accessible via HTTPS**."
https://helpdesk.ssls.com/hc/en-us/articles/206957109-How-can-I-complete-the-domain-control-validation-DCV-for-my-SSL-certificate-
https://features.cpanel.net/topic/autossl-dcv-validtion-by-https-too

LeoColomb:
Thanks for these details @westhouseit, appreciated! 👍 
I'm wondering if these additions are not too specific for some really isolated cases. Not everyone use Let's encrypt webroot validation! 😉
Also, are you sure these endpoints are durable enough to be considered as a standard or even as a good practice?

Thoughts @XhmikosR?

westhouseit:
Good point Leo, though I imagine a lot use cPanel AutoSSL. Here's the latest threads I could find:
https://forums.cpanel.net/threads/autossl-errors-due-to-http-redirects-in-cpanel-v68-0-despite-domain-validation-exclusion-conditions.621503/
https://forums.cpanel.net/threads/autossl-errors-not-renewing-certificate.630015/

westhouseit:
IETF reference for acme-challenge: https://tools.ietf.org/html/draft-ietf-acme-acme-04

westhouseit:
"pki-validation" and "acme-challenge" are listed as well-known: https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml

XhmikosR:
Hmm, I see the use case for sure, and I remember hitting this a couple of years ago with CPanel where I reported it to them and they added a workaround back then.

That being said, it's not a bad idea to have this, although it's a very specific use-case. Also, it definitely needs more comments with links and explanation.

Perhaps it should be off by default for starters?

LeoColomb:
@westhouseit Thanks for all these official links.
@XhmikosR suggestion to add them but keep them commented out by default seems the best way to do.
Could you do that?
Also, you must add all these relevant links as local documentation.
See [the point (1) here](https://github.com/h5bp/server-configs-apache/blob/ea022f440ed4c50dbfa0acc6194c06ddefcbc666/src/rewrites/rewrite_nowww.conf) as an example how to do.

Thanks again!

westhouseit:
@LeoColomb done, how's that?

src/rewrites/rewrite_http_to_https.conf
```diff
@@ -6,8 +6,20 @@
...
+#
+#     https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml
+#     https://tools.ietf.org/html/draft-ietf-acme-acme-12
+
 <IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{HTTPS} !=on
+   # (1)
+   # RewriteCond %{REQUEST_URI} !^/\.well-known/acme-challenge/[0-9a-zA-Z_-]+$
+   # RewriteCond %{REQUEST_URI} !^/\.well-known/cpanel-dcv/[0-9a-zA-Z_-]+$
```
LeoColomb: Are you sure these both regular expressions cannot be simplified?
Maybe `[\d-]` instead of `[0-9a-zA-Z_-]`? (I'm not sure that said...)

westhouseit:
I think you meant `[\w-]` right? Also, I missed that the acme-challenge regex has changed since the docs allow any base64url for the token: https://tools.ietf.org/html/draft-ietf-acme-acme-12#section-8.4

LeoColomb:
> I think you meant [\w-] right?

Woops, yes, sorry! 😇 

> I missed that the acme-challenge regex

Any news? 🙂

PR #165, opened 2018-11-21
Update pre-compressed_content_brotli.conf

Malvoz:
**Brotli requires https**, adding a note on that.

Indications to back up that statement:
https://groups.google.com/forum/#!topic/brotli/fN6M9A8aRnM
https://www.chromestatus.com/feature/5420797577396224

If this PR is merged, do I also have to add this to `dist/.htaccess` or is that built from the source files?

### DIFF
--- a/src/web_performance/pre-compressed_content_brotli.conf
+++ b/src/web_performance/pre-compressed_content_brotli.conf
@@ -5,9 +5,11 @@
 # Serve brotli compressed CSS, JS, HTML, SVG, ICS and JSON files
 # if they exist and if the client accepts br encoding.
 #
-# (!) To make this part relevant, you need to generate encoded
-# files by your own. Enabling this part will not auto-generate
-# brotlied files.
+# (!) To make this part relevant, you need to:
+# 1. Serve the website over a secure connection (Brotli requires 
+# `https://`).
+# 2. Generate encoded files by your own. Enabling this part will not 
+# auto-generate brotlied files.
 #
 # https://httpd.apache.org/docs/current/mod/mod_brotli.html#precompressed
 


LeoColomb:
Thanks for your suggestion @Malvoz!

I'm concerned by the following points:
* The chrome status ticket doesn't really say Brotli is exclusively served on HTTPS;
* The [chromium source](https://cs.chromium.org/chromium/src/net/url_request/url_request_http_job.cc?q=brotli+accept-encoding&sq=package:chromium&dr=C&l=659) include at least localhost as a Brotli-compatible environment;
* Anyway the constraint is when _fetching_ ressource, not when _serving_ them.

Malvoz:
@LeoColomb I couldn't read that source code if I tried 😭 

I don't understand the fetch vs serve scenario you're referring to. 

This tweet:
https://twitter.com/igrigorik/status/690230711924617216

or Yoav's blog post:
https://medium.com/@yoavweiss/well-the-technical-reason-for-brotli-being-https-only-is-that-otherwise-there-s-a-very-high-508f15f0ad95

Seems to indeed suggest resources (maybe from intermediate proxies) over an insecure protocol wont be able to serve Brotli pre-compressed content.

LeoColomb:
> I don't understand the fetch vs serve scenario you're referring to.

I mean that secure condition comes from some clients, not from the server.  
If you mount a server without secure connexion and request Brotli ressources directly with a raw client (let's say curl), everything will work as expected. 
In other words, the secure environment is not a requirement to make the server answer correctly and standard-compliantly.

LeoColomb:
Merged with commit 197eff471d9c64710a4e0382c1a0a8da3ff6fb85

I've edited you addition to switch from a requirement warning to a note.
Thanks for your contribution! 🎉

PR #168, opened 2018-11-24
Add SSL conf files

### LINKED ISSUE #72: Add SSL section
It would be fantastic to have an optimised/secured SSL section:
- disable insecure cyphers
- disable insecure protos
- enable/disable renegotiation

and so on ..

LeoColomb:
Fix #72
Prepare #1

---
Some comments may need to be improved, suggestions?


XhmikosR:
Not sure I get the conf file preparation. Will we distribute conf and the .htaccess file?

LeoColomb:
Exactly.
* Keep `.htaccess` which becomes slowly a release bundle (still 100% valid)
* Allow conf files since more and more people may use config with root-access machine, for instance Docker containers (plus all arguments in #1).

It will also allow us to rewrite tests (I'm on it) to increase coverage and get more precision on _what_ to test.

LeoColomb:
Thanks a lot for these fixes! 😍

XhmikosR:
NP, feel free to squash those into one commit, I just did them via the web UI.

LeoColomb:
Merged into `feat/conf-repo`, #188.

PR #169, opened 2018-11-25
Update trace_method.conf

Malvoz:
For consistency - similarly to [`ServerTokens Prod`](https://github.com/h5bp/server-configs-apache/blob/master/src/security/server_software_information.conf) which is only available from the main server config file, the `TraceEnable` directive shouldn't be enabled by default. 

This PR adds an alternative method for disallowing `TRACE` from `.htaccess` instead, which is enabled by default.

### DIFF
--- a/src/security/trace_method.conf
+++ b/src/security/trace_method.conf
@@ -10,26 +10,17 @@
 # Modern browsers now prevent TRACE requests being made via JavaScript,
 # however, other ways of sending TRACE requests with browsers have been
 # discovered, such as using Java.
-
-# Disallow TRACE HTTP Method:
 #
 # (!) If you have access to the main server configuration file, use the 
-# `TraceEnable` directive instead.
-
-<IfModule mod_rewrite.c>
-    RewriteEngine On
-    RewriteCond %{REQUEST_METHOD} ^TRACE [NC]
-    RewriteRule .* - [R=405,L]
-</IfModule>
-
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-
-# (!) The `TraceEnable` directive will only work in the main server
-# configuration file, so don't try to enable it in the `.htaccess` file!
+#     `TraceEnable` directive instead.
 #
 # https://tools.ietf.org/html/rfc7231#section-4.3.8
 # https://www.owasp.org/index.php/Cross_Site_Tracing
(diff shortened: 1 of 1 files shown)


XhmikosR:
I can't comment on this being I have no experience with this specific case.

Malvoz:
This is merely a PR to remain consistent in not assuming the developer has access to the main server config file. The [`TraceEnable`](https://httpd.apache.org/docs/2.4/mod/core.html#traceenable) directive is not available in the context of `.htaccess`.

LeoColomb:
Thanks for your PR, @Malvoz!

> the reference to the `ServerTokens Prod`

Well, that's not exactly the same since `ServerTokens Prod` & `ServerSignature Off` don't do the exact same thing.
Here, the two parts _do_ the same thing, right?

> alternative method for disallowing `TRACE` from `.htaccess`

Did you actually test it?

Malvoz:
A bit confusing to me is that `TraceEnable` is disabled (commented out) in the /dist version but not in /src whereas `ServerTokens Prod` is disabled in both places. So in the dist version both `TraceEnable` and `ServerTokens Prod` are disabled which makes sense (now that I realized it).

> Did you actually test it?

I thought these things were also tested before implemented on your side. No I did not test the `TRACE` Method, however I did test the same snippet replacing `TRACE` with `GET` and successfully returned a 405 Method Not Allowed status for the `GET` request.

Malvoz:
> Did you actually test it?

I thought these things were also tested before implemented on your side. No I did not test the `TRACE` Method, however I did test the same snippet replacing `TRACE` with `GET` and successfully returned a 405 Method Not Allowed status for the `GET` request.

LeoColomb:
@Malvoz Ooops, sorry for late reply, I didn't see your reply.

> A bit confusing to me is that `TraceEnable` is disabled

This is because building configuration: https://github.com/h5bp/server-configs-apache/blob/029d2447a784180d874213c8ac1a9cc814d7599c/htaccess.conf#L78

> I thought these things were also tested before implemented on your side.

Yes, it is partly tested manually. But in this case I wonder if a request with TRACE method will even make Apache read the `.htaccess`. [It seems to](https://stackoverflow.com/a/18960026), but that's what lead to my question. 🙂

LeoColomb:
Thanks, @Malvoz.
I'm moving the "conf" part in #188, making your suggestion as the preference for `.htaccess` files.

PR #170, opened 2018-11-27
Support hashed asset names in cache-busting

LeoColomb:


src/web_performance/filename-based_cache_busting.conf
```diff
@@ -13,5 +13,5 @@
 <IfModule mod_rewrite.c>
     RewriteEngine On
     RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^(.+)\.(\d+)\.(bmp|css|cur|gif|ico|jpe?g|m?js|png|svgz?|webp|webmanifest)$ $1.$3 [L]
+    RewriteRule ^(.+)\.(\w+)\.(bmp|css|cur|gif|ico|jpe?g|m?js|png|svgz?|webp|webmanifest)$ $1.$3 [L]
```
XhmikosR: You should allow `-` too.

XhmikosR: I'm probably wrong in the sense it's not common, so scratch my comment :)

PR #166, opened 2018-11-23
Setting headers based on content-type

LeoColomb:
Setting headers based on content-type instead of `FilesMatch` conditions.

~~WIP: need performance tests~~

fix #163

PR #176, opened 2019-02-12
Use server-configs-test for tests

### LINKED ISSUE #45: Clean up and improve tests

LeoColomb:
Switch from local Node.js based unit tests to k6 based generic tests.

This change comes with following improvements:
* More accurate HTTP requests/tests and better output*
* Will help increasing coverage (testing redirection, proto, tls, etc.) 🎉 
* Will help to switch smoothly to a conf repo (#1) 🎉 
  * Will help testing additions like #168
  * Should help resolving case like #166
  * Will enforce compatibility with httpd Docker image
* Less maintenance (especially npm dependencies)
* Reduce file pollution (basically fixtures)

Fix #45 

---
\* I don't know why but colors in the console output are missing... (it [has color on server-configs-nginx builds](https://travis-ci.org/h5bp/server-configs-nginx/builds/492124598), so really, idk why)


XhmikosR:
Wow, this is a huge change! Not sure I can review though, apart from the fact that it seems to work.

LeoColomb:
@XhmikosR Sorry, for this monster change! 😇
But actually, if you exclude fixtures (`test/fixtures/**`) only 20 files are involved and ~half of them are just moves/renames (build tests are now in `test/build/` instead of root `test/`).
The most important change is for `.travis-ci.yml`.

XhmikosR:
If it works, it looks good enough to me :)

bin/build.sh
```diff
@@ -13,11 +13,7 @@ create_htaccess() {
     local file="${1}"
     local config="${2}"
 
-    local version
-    version=$(grep version < "${repo_root}/package.json" | \
-        head -1 | awk -F: '{ print $2 }' | sed 's/[",\t ]//g')
-
-    insert_line "# Apache Server Configs v$version | MIT License" "$file"
+    insert_line "# Apache Server Configs | MIT License" "$file"
```
XhmikosR: Is the version removal intentional here?

LeoColomb: 50% intentional, for now...
I'll try to re-add it later, but without the need for parsing `package.json`. Maybe only for the release distribution also.

LeoColomb: Actually, yes, it may be better to re-add it now. I'll do this.

XhmikosR: Since you removed package.json because it's unused, we need another way to track this.

BTW you shouldn't remove package.json IMO. Don't we publish this to npm?

LeoColomb: Well, I see many people complaining about using npm just for downloading a single file (not specifically for this project, but same), even more for project that do not relate to javascript/Node.js ecosystem.

What do you think about using GitHub' release assets for distribution?

It was relevant to have a `package.json` as the project required some npm dependencies for testing, with this PR this no more the case.

If we still need to publish version to npm, we can add a small package for downloading the file from the release assets (as many other vendors do for this static file distribution situation).

Right? 🙂

XhmikosR: I don't agree. If we publish to npm, it's the right way to do it. Without any deps or anything, a simple package.json.

LeoColomb: Alright! 😉 Let me re-add a `package.json`.

XhmikosR: Let me add it and rebase the branch also

XhmikosR: It seems you have removed the dist/.htaccess file too. Is that intentional? If so it should be added in .gitignore otherwise reverted.

LeoColomb: > it should be added in .gitignore

Good catch, added.

LeoColomb: Thanks for the `package.json`!

XhmikosR: So, how are the npm deployments going to be made? Because now one will have to build this thing in order to do `npm publish`. I can't being I'm on Windows for example.

LeoColomb: Ideally with CI automation (or using GitHub Actions 😇). The `.htaccess` is still being deployed to the release on GitHub.

Do you prefer I extract these changes to open another PR? I have to admit, it's not really related to the tests themselves.

XhmikosR: I don't have any preference as long as npm publishing works.

LeoColomb: For sure. I added npm as a deploy provider when Travis-CI builds on tags (f1943a5). Let me know if it's OK for you.

(Fyi, I used to build with WSL for previous npm publishing. 😉)

bin/build.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+declare version="3.1.0";
```
XhmikosR: This should be retrieved from package.json so that there's only one place the version needs to change.

.travis.yml
```diff
@@ -1,37 +1,62 @@
...
+before_script:
+  - bash -c "./bin/build.sh test/fixtures/.htaccess test/build/htaccess_fixture.conf"
+  - docker pull httpd
+  - |
+    docker run -d -p 80:80 -p 443:443 --name server \
+      -v $TRAVIS_BUILD_DIR/test/fixtures:/usr/local/apache2/htdocs \
+      -v $TRAVIS_BUILD_DIR/test/vhosts/httpd.conf:/usr/local/apache2/conf/httpd.conf \
+      httpd
+
+#      -v $TRAVIS_BUILD_DIR/test/vhosts:/usr/local/apache2/vhosts \
```
XhmikosR: Maybe remove the commented out lines here?

bin/build.sh
LeoColomb: @XhmikosR You're too fast for me! ^^
Did it.

.travis.yml
LeoColomb: I will restore them later but I can cleanup for now, sure.

XhmikosR:
One thing, shouldn't we have dist/.htaccess to the repo?

PR #181, opened 2019-03-18
[CSP] enhancements to example policy

### LINKED ISSUE #156: [CSP] enhancements to default policy
The current default: `Header set Content-Security-Policy "script-src 'self'; object-src 'self'"` means authors need to explicitly specify all types of directives that apply which may be a bit tedious to set up and maintain, e.g. (not considering crossorigin resources) a sites policy could eventually look something like:
```
Header set Content-Security-Policy
  "base-uri 'none';
   connect-src 'self';
   font-src 'self'; 
   form-action 'self';
   frame-ancestors 'none';
   img-src 'self'; 
   manifest-src 'self'; 
   media-src 'self';
   object-src 'self'; 
   prefetch-src 'self';
   style-src 'self'; 
   script-src 'self'; 
   worker-src 'self';
   upgrade-secure-requests"
```

Instead I suggest setting `default-src 'self'` which acts like a fallback src to all [_fetch_](https://www.w3.org/TR/CSP/#directives-fetch) directives.

With this approach, you would still need to include: 

- the `base-uri` [document directive](https://www.w3.org/TR/CSP/#directives-document).
- the `form-action` and `frame-ancestors` [suggested in [#154](https://github.com/h5bp/server-configs-apache/issues/154)] [navigation directives](https://www.w3.org/TR/CSP/#directives-navigation).

This is important because unlike fetch directives, the **document**- and **navigation** directives **do not fallback** to `default-src`, and allows all origins by default.

### New example default policy might look like:

```
Header set Content-Security-Policy
  "base-uri 'none'; #[1]
   default-src 'self'; 
   form-action 'self'; 
   frame-ancestors 'none';
   upgrade-insecure-requests" #[2]
```

[1] Note that setting `base-uri` to `'self'` while not intending to use the [`<base>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) could break site functionality if an attacker manages to inject a `<base href="https://my-site.com">`. Because the majority of sites do not use the base element, the value `'none'` may be a better default.

[2] Additionally, the [`upgrade-insecure-requests`](https://w3c.github.io/webappsec-upgrade-insecure-requests/) directive is recommended for secure context configurations to prevent mixed content.

<hr>

Finally, I think having an example of adding an external resource would be useful, such as allowing fonts from another origin.
`font-src` is a fetch directive and falls back to `default-src`, however we still need to specify `'self'` if we wish to add an external url such as google fonts:
```
font-src 'self' https://fonts.googleapis.com;
```

Malvoz:
Fix #156

Also, with these changes the CSP `frame-ancestors` directive description now mentions the relation to `X-Frame-Options`  (https://w3c.github.io/webappsec-csp/#frame-ancestors-and-frame-options). Seems logical `X-Frame-Options` should say something about `frame-ancestors`/CSP? I can probably do that tomorrow, really tired atm. :)

### DIFF
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -8,18 +8,69 @@
 # This can be done by setting a `Content Security Policy` which
 # whitelists trusted sources of content for your website.
 #
-# The example header below allows ONLY scripts that are loaded from
-# the current website's origin (no inline scripts, no CDN, etc).
-# That almost certainly won't work as-is for your website!
+# There is no one policy that fits all websites, you will have to modify
+# the `Content-Security-Policy` directives in the example below depending
+# on your needs.
 #
-# To make things easier, you can use an online CSP header generator
-# such as: https://www.cspisawesome.com/.
+# The example policy below ensures that:
 #
+#  (1) The `<base>` element is not allowed on the website. This is to
+#      prevent attackers from changing the locations of resources loaded
+#      from relative URLs.
+#
+#      If you want to use the `<base>` element, then `base-uri 'self'`
+#      can be used instead.
+#
+#  (2) All resources are restricted to the origin of the current website
+#      by setting the `default-src` directive to `'self'` - which acts as a
+#      fallback to all "Fetch directives" (https://developer.mozilla.org/en-US/docs/Glossary/Fetch_directive).
+#
+#      This is conveniant as you do not have to specify all Fetch directives
+#      that apply to your site, for example:
+#      `connect-src 'self'; font-src 'self'; script-src 'self'; style-src 'self'`, etc.
+#
+#      This restriction also means that you must explicitly define from
+#      which site(s) your website is allowed to load resources from.
+#
+#  (3) Scripts are only allowed to be loaded from the current website
+#      and an example CDN website.
+#
+#  (4) Form submissions are only allowed from the current website by
+#      setting: `form-action 'self'`.
+#
+#  (5) Prevents other websites to embed your website with e.g. the
+#      `<iframe>` or `<object>` element, by setting `frame-ancestors 'self'`.
+#
+#	   The `frame-ancestors` directive helps avoid "Clickjacking" attacks
+#      and is similar to the `X-Frame-Options` header.
+#
+#      Browsers that support the CSP header will ignore `X-Frame-Options`
+#      if `frame-ancestors` is also specified.
+#
+#  (6) Forces the browser to treat all the resources that are served over
+#      HTTP as if they were loaded securely over HTTPS by setting the
+#      `upgrade-insecure-requests` directive.
+#
+#      Please note that `upgrade-insecure-requests` does not ensure
+#      HTTPS for the top-level navigation. If you want to force the
+#      website itself to be loaded over HTTPS you must include the
+#      `Strict-Transport-Security` header.
+#
+# To make your CSP implementation easier, you can use an online CSP header
+# generator such as:
+# https://www.cspisawesome.com/
+# https://report-uri.com/home/generate/.
+#
+# It is encouraged that you validate your CSP header using a CSP validator
+# such as:
+# https://csp-evaluator.withgoogle.com
+#
+# https://csp.withgoogle.com/docs/
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
-# https://www.w3.org/TR/CSP3/
-# https://content-security-policy.com/
 # https://www.html5rocks.com/en/tutorials/security/content-security-policy/
+# https://www.w3.org/TR/CSP/
 
 <IfModule mod_headers.c>
-    Header set Content-Security-Policy "script-src 'self'; object-src 'self'" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
+    #                                   (1)              (2)                 (3)                                        (4)                 (5)                     (6)
+    Header set Content-Security-Policy "base-uri 'none'; default-src 'self'; script-src 'self' https://cdn.example.com; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
 </IfModule>
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -18,7 +18,7 @@
 #      by setting the `default-src` directive to `'self'` - which acts as a
 #      fallback to all "Fetch directives" (https://developer.mozilla.org/en-US/docs/Glossary/Fetch_directive).
 #
-#      This is conveniant as you do not have to specify all Fetch directives
+#      This is convenient as you do not have to specify all Fetch directives
 #      that apply to your site, for example:
 #      `connect-src 'self'; font-src 'self'; script-src 'self'; style-src 'self'`, etc.
 #
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -8,7 +8,7 @@
 # This can be done by setting a `Content Security Policy` which
 # whitelists trusted sources of content for your website.
 #
-# There is no one policy that fits all websites, you will have to modify
+# There is no policy that fits all websites, you will have to modify
 # the `Content-Security-Policy` directives in the example below depending
 # on your needs.
 #
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -58,7 +58,6 @@
 #
 # To make your CSP implementation easier, you can use an online CSP header
 # generator such as:
-# https://www.cspisawesome.com/
 # https://report-uri.com/home/generate/
 #
 # It is encouraged that you validate your CSP header using a CSP validator
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -14,7 +14,7 @@
 #
 # The example policy below aims to:
 #
-#  (1) Restrict all fetches by default from the origin of the current website
+#  (1) Restrict all fetches by default to the origin of the current website
 #      by setting the `default-src` directive to `'self'` - which acts as a
 #      fallback to all "Fetch directives" (https://developer.mozilla.org/en-US/docs/Glossary/Fetch_directive).
 #
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -59,7 +59,7 @@
 # To make your CSP implementation easier, you can use an online CSP header
 # generator such as:
 # https://www.cspisawesome.com/
-# https://report-uri.com/home/generate/.
+# https://report-uri.com/home/generate/
 #
 # It is encouraged that you validate your CSP header using a CSP validator
 # such as:
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -12,9 +12,9 @@
 # the `Content-Security-Policy` directives in the example below depending
 # on your needs.
 #
-# The example policy below ensures that:
+# The example policy below aims to:
 #
-#  (1) All resources are restricted to the origin of the current website
+#  (1) Restrict all fetches by default from the origin of the current website
 #      by setting the `default-src` directive to `'self'` - which acts as a
 #      fallback to all "Fetch directives" (https://developer.mozilla.org/en-US/docs/Glossary/Fetch_directive).
 #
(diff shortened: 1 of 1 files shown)


src/security/content-security-policy.conf
```diff
@@ -8,18 +8,69 @@
...
+# It is encouraged that you validate your CSP header using a CSP validator
+# such as:
+# https://csp-evaluator.withgoogle.com
+#
+# https://csp.withgoogle.com/docs/
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
-# https://www.w3.org/TR/CSP3/
-# https://content-security-policy.com/
 # https://www.html5rocks.com/en/tutorials/security/content-security-policy/
+# https://www.w3.org/TR/CSP/
```
LeoColomb: 👍

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,69 @@
...
+#      `upgrade-insecure-requests` directive.
+#
+#      Please note that `upgrade-insecure-requests` does not ensure 
+#      HTTPS for the top-level navigation. If you want to force the
+#      website itself to be loaded over HTTPS you must include the
+#      `Strict-Transport-Security` header.
+#
+# To make your CSP implementation easier, you can use an online CSP header 
+# generator such as: 
+# https://www.cspisawesome.com/
```
LeoColomb: 👎 `cspisawesome.com` is very outdated, still using CSP1.

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,69 @@
...
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
-# https://www.w3.org/TR/CSP3/
-# https://content-security-policy.com/
 # https://www.html5rocks.com/en/tutorials/security/content-security-policy/
+# https://www.w3.org/TR/CSP/
 
 <IfModule mod_headers.c>
-    Header set Content-Security-Policy "script-src 'self'; object-src 'self'" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
+    #                                   (1)              (2)                 (3)                                        (4)                 (5)                     (6)
+    Header set Content-Security-Policy "base-uri 'none'; default-src 'self'; script-src 'self' https://cdn.example.com; form-action 'self'; frame-ancestors 'self'; upgrade-insecure-requests" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
```
LeoColomb: * `default-src` first.
* `https://cdn.example.com` is here for the example of course, but I don't like adding weakness for the examples, even if nothing is behind `cdn.example.com`. Also using hosts for `script-src` is not recommended.
* `form-action` duplicates the `default-src`, keeping defaults as default is sufficient. 😉 
* `frame-ancestors` should be set to `none` to match with current h5bp value.

Malvoz: > default-src first.

Done!

> Also using hosts for `script-src` is not recommended.

I don't follow, if you want to include a script from another origin then this is the method. <br>However:

> `https://cdn.example.com` is here for the example of course, but I don't like adding weakness for the examples [...]

Removed!

> `frame-ancestors` should be set to `none` to match with current h5bp value. 

Done!

> `form-action` duplicates the `default-src` [...]

`form-action` is a _navigation_ directive, only _fetch_ directives falls back to `default-src`. The default value of `form-action` is `*`.

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,69 @@
...
-# The example header below allows ONLY scripts that are loaded from
-# the current website's origin (no inline scripts, no CDN, etc).
-# That almost certainly won't work as-is for your website!
+# There is no one policy that fits all websites, you will have to modify 
+# the `Content-Security-Policy` directives in the example below depending
+# on your needs.
 #
-# To make things easier, you can use an online CSP header generator
-# such as: https://www.cspisawesome.com/.
+# The example policy below ensures that:
```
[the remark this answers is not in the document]
Malvoz: I'm sorry - I don't understand, do you want me to be more or less detailed about things?...

src/security/content-security-policy.conf
LeoColomb: > I don't follow, if you want to include a script from another origin then this is the method.

I was thinking the same thing two days before, but:

> Host whitelists can frequently be bypassed. Consider using 'strict-dynamic' in combination with CSP nonces or hashes.
[csp-evaluator.withgoogle.com](https://csp-evaluator.withgoogle.com/)

That said I didn't verify the source of that displayed info.

> `form-action` is a [_navigation_ directive](https://www.w3.org/TR/CSP/#directives-navigation)

Oh, you're totally right, sorry.

Malvoz: > Host whitelists can frequently be bypassed. Consider using `'strict-dynamic'` in combination with CSP nonces or hashes.

This blog post (I didn't read it all yet): https://websec.be/blog/cspstrictdynamic/ seems to go into some detail on insecure whitelists, the main issue seems to be allowing endpoints that serve JSONP. 

`script-src` is easier to implement but it seems you're right. `script-src` is removed from the example policy however.

Malvoz:
> `form-action` is a [navigation directive](https://www.w3.org/TR/CSP/#directives-navigation), only fetch directives falls back to [`default-src`](https://www.w3.org/TR/CSP/#directive-default-src).

Do you want me to clarify that, here(?):

https://github.com/h5bp/server-configs-apache/blob/a3976540c91b8c7a8e20a39d66138c0d08826e83/src/security/content-security-policy.conf#L35-L36

src/security/content-security-policy.conf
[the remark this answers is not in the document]
LeoColomb: Sorry, I was not clear. Yes, I think this is a bit too verbose. But let's keep it as it is for now.

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,66 @@
...
-# That almost certainly won't work as-is for your website!
+# There is no one policy that fits all websites, you will have to modify
+# the `Content-Security-Policy` directives in the example below depending
+# on your needs.
 #
-# To make things easier, you can use an online CSP header generator
-# such as: https://www.cspisawesome.com/.
+# The example policy below ensures that:
 #
+#  (1) All resources are restricted to the origin of the current website
```
XhmikosR: I don't think that's true. There are directives that don't inherit from `default-src`.

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,66 @@
...
+# It is encouraged that you validate your CSP header using a CSP validator
+# such as:
+# https://csp-evaluator.withgoogle.com
+#
+# https://csp.withgoogle.com/docs/
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
-# https://www.w3.org/TR/CSP3/
-# https://content-security-policy.com/
 # https://www.html5rocks.com/en/tutorials/security/content-security-policy/
+# https://www.w3.org/TR/CSP/
```
XhmikosR: I'd definitely add @ScottHelme's https://report-uri.com/home/generate and even @marumari's https://addons.mozilla.org/en-US/firefox/addon/laboratory-by-mozilla/ addon

src/security/content-security-policy.conf
XhmikosR: It is true for the directives you use, but generally it's not true.

src/security/content-security-policy.conf
LeoColomb: @XhmikosR The report-uri.com link is already here, few lines above.
I'm not a big fan of browser extensions/addons (installation steps, browser specificity, etc.)...

XhmikosR: Np, all good then. I also missed the reporturi link :P

src/security/content-security-policy.conf
LeoColomb: Indeed. That's also why [I suggested](https://github.com/h5bp/server-configs-apache/pull/181#discussion_r266406669) to reduce verbosity of this documentation and prefer linking to official and rigorous docs already listed. What do you think?

Malvoz: @XhmikosR @LeoColomb 

I've updated for less verbosity to address both those statements (also cleaned up indentation) in https://github.com/h5bp/server-configs-apache/pull/181/commits/14b3aa98dd12bbf105981ad35a09600d2cadaf4d.

Thanks for the input, don't hesitate to tell me if I missed something.

src/security/content-security-policy.conf
```diff
@@ -8,18 +8,66 @@
...
 
 <IfModule mod_headers.c>
-    Header set Content-Security-Policy "script-src 'self'; object-src 'self'" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
+    #                                   (1)                 (2)              (3)                 (4)                     (5)
+    Header set Content-Security-Policy "default-src 'self'; base-uri 'none'; form-action 'self'; frame-ancestors 'none'; upgrade-insecure-requests" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
```
LeoColomb: Suggestion from https://github.com/h5bp/server-configs-nginx/pull/224#issuecomment-476101646:

```suggestion
    Header set Content-Security-Policy "default-src 'none'; frame-ancestors 'none'; form-action 'none'; upgrade-insecure-requests 1; block-all-mixed-content; disown-opener; sandbox; reflected-xss block; base-uri 'none'; referrer no-referrer-when-downgrade" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
```

Basically:
* Shouldn't we use `none` for `default-src`? This might be to restrictive for scripts and styles, but more secure for all others.
* `block-all-mixed-content`, `disown-opener` and `sandbox` seem safe enough but are they useful?
* `reflected-xss block` modern way to [`X-XSS-Protection`](https://github.com/h5bp/server-configs-apache/blob/e12ae05c4ae690e36386de0efc2cfc0186f2f4f2/src/security/x-xss-protection.conf).
* `referrer no-referrer-when-downgrade` alternative to [`Referrer-Policy`](https://github.com/h5bp/server-configs-apache/blob/e12ae05c4ae690e36386de0efc2cfc0186f2f4f2/src/security/referrer-policy.conf).

Thoughts?

XhmikosR: I think we should keep this simple, at least for starters.

So I wouldn't duplicate the directives that are covered by others like Referrer-Policy etc.

About default-src, while I agree, it's the issue we discussed. Many other CSP directives will inherit this. Perhaps `'self'` is a more sane default.

About block-all-mixed-content, disown-opener and sandbox, not sure I'd add them right now. But this should be revisited later for sure.

Malvoz: I agree with @XhmikosR keeping `default-src` to `'self'` mainly because it is an easier implementation model than specifying each and every fetch directive used by the application.

`sandbox` is very restrictive and developers must be aware of which Web APIs are being used in their application then allow them accordingly. Many of these [values](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox#Syntax) would have to be enabled again anyways to preserve normal web behaviors, such as `allow-forms`, `allow-scripts`, `allow-top-navigation`. So just having other CSP directives allowing _who_ can use things instead of `sandbox` dictating _what_ seems like the best default. Developers who want advanced and super strict policies certainly will look closer at directives such as `sandbox`.

`disown-opener` seems to be deprecated: https://github.com/w3c/webappsec-csp/issues/194. I think a new HTTP header [`Cross-Origin-Opener-Policy`](https://github.com/whatwg/html/issues/3740) is to replace it.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/referrer), `referrer` is obsolete and we should stick to the `Referrer-Policy` header instead.

`block-all-mixed-content` is similar to `upgrade-insecure-requests` and if the latter is specified then the `block-all-mixed-content` directive is ignored. Although `block-all-mixed-content` has a tiny bit wider support I think we should default to `upgrade-insecure-requests` which [Google describes](https://developers.google.com/web/fundamentals/security/prevent-mixed-content/fixing-mixed-content#upgrading_insecure_requests) as "the newest and best tools to automatically fix mixed content" - the directive simply upgrades HTTP to HTTPS while `block-all-mixed-content` (seems to me) to straight up block HTTP requests and Google describe that directive with: "The downside of using `block-all-mixed-content` is, perhaps obviously, that all content is blocked."

> `reflected-xss block` modern way to [`X-XSS-Protection`](https://github.com/h5bp/server-configs-apache/blob/e12ae05c4ae690e36386de0efc2cfc0186f2f4f2/src/security/x-xss-protection.conf).

I don't know what this is, I've not been able to find such a CSP directive defined anywhere.

LeoColomb: Thanks a lot for feedbacks!
And well done, @Malvoz, very clear 👍 

> > `reflected-xss block` modern way to [`X-XSS-Protection`](https://github.com/h5bp/server-configs-apache/blob/e12ae05c4ae690e36386de0efc2cfc0186f2f4f2/src/security/x-xss-protection.conf).
> 
> I don't know what this is

Seems deprecated too: https://www.chromestatus.com/feature/5769374145183744

All good then!

PR #184, opened 2019-03-24
Always unset X-Powered-By header

### LINKED ISSUE #183: Directive "Header unset X-Powered-By" is not respected
Hey there, 
just as title, noting also that `mod_headers.c` _is_ enabled, and it's happening on a shared hosting server where I have no access to the PHP ini.  
After a couple testing and some fortunate google-fu, I was able to finally enforce it adding `always` to the mix

    Header always unset X-Powered-By

I can't really understand why the directive it is not respected without it, also there are no redirects in the request. 

In the [docs ](https://httpd.apache.org/docs/current/mod/mod_headers.html) I wasn't able to find much about possible side effects for this option.  

So I'm wondering if it could be a sensed default to insert for this specific rule, or maybe put a notice in the comments suggesting this approach if the directive is not working.

I can make a PR if requested.

Thank you!

GlassGruber:
as per #183, `always` can be used to ensure the directive is always applied. In this case should be a sensed default. Also as found [here](https://www.tunetheweb.com/security/http-security-headers/server-header/) a good default is to use both versions since the `always` option is reported as not always working in some server configurations.

Fix #183

### DIFF
--- a/src/security/x-powered-by.conf
+++ b/src/security/x-powered-by.conf
@@ -22,7 +22,4 @@
 # Note you need both below as the "always" option doesn't always work with some servers for some reasons.
 <IfModule mod_headers.c>
     Header always unset X-Powered-By
-    Header unset X-Powered-By  
 </IfModule>
-
-
--- a/src/security/x-powered-by.conf
+++ b/src/security/x-powered-by.conf
@@ -18,8 +18,6 @@
 #
 # https://php.net/manual/en/ini.core.php#ini.expose-php
 
-
-# Note you need both below as the "always" option doesn't always work with some servers for some reasons.
 <IfModule mod_headers.c>
     Header always unset X-Powered-By
 </IfModule>
--- a/src/security/x-powered-by.conf
+++ b/src/security/x-powered-by.conf
@@ -18,6 +18,11 @@
 #
 # https://php.net/manual/en/ini.core.php#ini.expose-php
 
+
+# Note you need both below as the "always" option doesn't always work with some servers for some reasons.
 <IfModule mod_headers.c>
-    Header unset X-Powered-By
+    Header always unset X-Powered-By
+    Header unset X-Powered-By  
 </IfModule>
+
+


src/security/x-powered-by.conf
```diff
@@ -18,6 +18,11 @@
...
 
+
+# Note you need both below as the "always" option doesn't always work with some servers for some reasons.
 <IfModule mod_headers.c>
-    Header unset X-Powered-By
+    Header always unset X-Powered-By
+    Header unset X-Powered-By  
 </IfModule>
+
+
```
LeoColomb: There are two extra blank lines added.

src/security/x-powered-by.conf
```diff
@@ -18,6 +18,11 @@
 #
 # https://php.net/manual/en/ini.core.php#ini.expose-php
 
+
+# Note you need both below as the "always" option doesn't always work with some servers for some reasons.
 <IfModule mod_headers.c>
-    Header unset X-Powered-By
+    Header always unset X-Powered-By
+    Header unset X-Powered-By
```
LeoColomb: Hmm, I prefer we stick to only one directive.
Not sure about the JBoss case, I can't find other references where the issue is mentioned, so the directive with the `always` condition is sufficient.

GlassGruber: Sure, I'm really no expert in this.

src/security/x-powered-by.conf
```diff
@@ -18,6 +18,8 @@
 #
 # https://php.net/manual/en/ini.core.php#ini.expose-php
 
+
+# Note you need both below as the "always" option doesn't always work with some servers for some reasons.
```
LeoColomb: This comment is no longer relevant, you can remove it as much as the line above. 😉

GlassGruber: Gah silly leftover, fixed. :P

XhmikosR:
Definitely makes sense, but now that I see this, I wonder if there are more places we should `always` too?

LeoColomb:
Indeed, maybe all `Header` directives in `src/security/` are eligible for the `always` condition. To be honest, the documentation isn't very clear how we should use `always`, so guessing where to place it is not trivial.

Malvoz:
I tested this on Apache 2.4.37, PHP 7.3.1 and `always unset` did not work while `unset` did. Just FYI :)

LeoColomb:
So @GlassGruber was right when keeping the `onsuccess` condition.
Fixed with fae84edd797c98e9221442a7560e362855506d7d.

Issue #153, opened 2018-05-16
Remove "ServerTokens Prod" setting


creopard:
This setting simply can't be set in .htaccess, but only in httpd.conf file according to Apache's specification:
http://httpd.apache.org/docs/2.2/en/mod/core.html#servertokens
Doing so will just result in an 500 internal error.

The spec's line "Context:" at "ServerTokens Directive" only lists "server config" but not ".htaccess"

LeoColomb:
That's why there is a warning, right? ;)

https://github.com/h5bp/server-configs-apache/blob/master/src/security/server_software_information.conf#L19

LeoColomb:
Honestly I don't know. [Related commits](https://github.com/h5bp/server-configs-apache/search?q=ServerTokens&type=Commits) are not very verbose.

That say this may help to resolve #1.

Malvoz:
I don't think h5bp wants to assume a user does or does not have access to the main server configuration file. The current approach informs users _with_ access to the main config file the availability of `ServerTokens`, and clearly states to not use it if you don't.

However (unrelated to `ServerTokens` but related to comments on main config file), the [comment](https://github.com/h5bp/server-configs-apache/blob/ea022f440ed4c50dbfa0acc6194c06ddefcbc666/dist/.htaccess#L4) about avoiding `.htacces`s should also recommend setting `AllowOverride` to not allow `.htaccess` because: 

> When AllowOverride is set to allow the use of .htaccess files, httpd will look in every directory for .htaccess files. Thus, permitting .htaccess files causes a performance hit, whether or not you actually even use them!

_https://httpd.apache.org/docs/2.4/howto/htaccess.html#when_

XhmikosR:
Yeah, I'd say this should stay as is for the aforementioned reasons.

LeoColomb:
Agreed, closing.
@Malvoz Feel free to open a PR to add this recommendation. 👍

Issue #185, opened 2019-03-25
`Cache-Control: no-transform` redundant w/ TLS?


Malvoz:
Is `cache-control: no-transform` redundant over HTTPS?

> **Better to focus on https** which cannot be modified in flight [...]

<i>https://meta.discourse.org/t/add-cache-control-no-transform-http-header/25152/3</i>

> Some mobile providers will compress or alter content, in particular images, to save bandwidth when re-transmitting content over cellular networks. This could break digital signatures in some cases. "no-transform" will prevent that (but again: **doesn't matter for SSL** [...]

<i>https://isc.sans.edu/forums/diary/The+Security+Impact+of+HTTP+Caching+Headers/17033/</i>

As suggested by some folks in their articles, "payload transformations" aren't possible over HTTPS. This sounds logical, but I am unsuccessful in confirming this.

In https://tools.ietf.org/html/rfc7234#section-5.2.1.6 I'm unable to traverse up to find information on payload transformation in relation to the HTTP/HTTPS protocol.

<hr>

If it is the case that `no-transform` is redundant over HTTPS, then this:

https://github.com/h5bp/server-configs-apache/blob/e12ae05c4ae690e36386de0efc2cfc0186f2f4f2/src/web_performance/content_transformation.conf#L21

Could be (or only/and with a note stating no need to use if HTTPS?):

```
Header merge Cache-Control "no-transform" "expr=%{HTTPS} != 'on'"
````

LeoColomb:
Good catch.
Actually, does the `no-transform` still useful?

Malvoz:
> (Also: https://csswizardry.com/2019/03/cache-control-for-civilians/#no-transform)

> If you’re running over HTTPS—which you should be—then intermediaries and proxies can’t modify payloads anyway, so no-transform would be ineffective.

Everyone seems to agree with that statement!

If it is still useful? H5BP doesn't default to TLS configs (which is debatable), but today more than not - developers use TLS for their webistes. Which renders `no-transform` useless as we've established in our _extensive_ reasearch. :)

LeoColomb:
> H5BP doesn't default to TLS configs

I plan to, while working on #1. I've done it already for nginx configs. But we can't enable them by default since it require certificates we don't have.

> our _extensive_ reasearch

😎 

---
@XhmikosR Thoughts?

LeoColomb:
[...]

PR #188, opened 2019-05-06
Server conf files

### LINKED ISSUE #1: Should this repo be .conf files?
This topic has come up in the past numerous times, some references:
- h5bp/html5-boilerplate#169
- h5bp/html5-boilerplate#876
- h5bp/server-configs#59

The use of an optimized`.htaccess` file is a great improvement compared to a standard apache install - but it is really just a stepping-stone solution to configuring apache appropriately.

It's kind of hard to cater for all uses but e.g. potential users of this repo include:
- frontend developers using html5-boilerplate
- users, wanting to optimize their site
- developers using html5-boilerplate as part of their project
- server-savvy users

Right now, the first two groups are catered for pretty well - you can drop in the `.htaccess` file into any project and it just works, and is an improvement. Perhaps they can't modify their apache config files, so it's the only solution that works for them  - though arguably shared hosting is less common than in the past, being replaced by vps servers, reducing the significance/relevance of this point.

The last two groups however are not catered for particularly well, a (general) developer using apache may be aware that they could avoid using `.htaccess` files but, through convenience of using the files provided probably won't. A user familiar with apache either won't use the repo at all (and these are the users who could provide the most benefit in terms of future contributions), or needs to spend some time integrating the provided rules into their existing config files.

Perhaps only the boilerplate repo should have/include the `.htaccess` file (generated by some build script).

One way or another, I feel that users should be guided to _the_ best way to use apache, not stopping just short of the finish line.

Comments?

### LINKED ISSUE #72: Add SSL section

### LINKED ISSUE #139: Block access to all hidden files and directories is not working outside .htaccess
There is a rule in server-configs-apache/src/security/file_access.conf, which tries to prevent access to hidden files/directories:

```
<IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{REQUEST_URI} "!(^|/)\.well-known/([^./]+./?)+$" [NC]
    RewriteCond %{SCRIPT_FILENAME} -d [OR]
    RewriteCond %{SCRIPT_FILENAME} -f
    RewriteRule "(^|/)\." - [F]
</IfModule>
```

That rule is only working within a .htaccess file, but using .htaccess is really slow. As the [documentation](http://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond) mentions, `SCRIPT_FILENAME` has constraints when using in context

> If used in per-server context (i.e., before the request is mapped to the filesystem) SCRIPT_FILENAME and REQUEST_FILENAME cannot contain the full local filesystem path since the path is unknown at this stage of processing.

This makes the rule unusable and "unsafe". The good thing is, that the [documentation](http://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond) suggests to solve this problem using a URL-based look-ahead feature:

> In order to obtain the full local filesystem path of the request in per-server context, use an URL-based look-ahead %{LA-U:REQUEST_FILENAME} to determine the final value of REQUEST_FILENAME.

There is another problem to this rule used in a non-htaccess context. Rewrite rules are not inherited per default (see [RewriteOptions](http://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriteoptions)). So if someone relies on it as a global security measure, but than adds a `<Directory>` or `<VirtualHost>`, files and directories are not protected. Maybe it makes sense to add an inherit option like `InheritDown` or `InheritDownBefore`. This option got introduces with Apache2 Version 2.4.8.

The new rule could look like this:

```
<IfModule mod_rewrite.c>
	RewriteEngine On
	<IfModule mod_version.c>
		<IfVersion >= 2.4.8>
			RewriteOptions InheritDown
		</IfVersion>
	</IfModule>
	RewriteCond %{REQUEST_URI} "!(^|/)\.well-known/([^./]+./?)+$" [NC]
	RewriteCond %{LA-U:REQUEST_FILENAME} -d [OR]
	RewriteCond %{LA-U:REQUEST_FILENAME} -f
	RewriteRule "(^|/)\." - [F]
</IfModule> 
```

LeoColomb:
Closes #1
Closes #72
Closes #139

To-Do:
- [x] Merge SSL conf #168 
- [x] Add redirection tests
- [x] Add vhosts file examples
- [x] Documentation
- [x] Cleanup (remove redondant parts like #153)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jobs:
             -v $TRAVIS_BUILD_DIR/src:/usr/local/apache2/h5bp \
             -v $TRAVIS_BUILD_DIR/httpd.conf:/usr/local/apache2/conf/httpd.conf \
             httpd
-      script: ./test/bin/k6 run test/lib/index.js -e TESTS=basic-file-access:cache-busting:custom-errors:forbidden-files:enforce-gzip:precompressed-files-gzip
+      script: ./test/bin/k6 run test/lib/index.js -e TESTS=basic-file-access:cache-busting:custom-errors:forbidden-files:enforce-gzip:precompressed-files-gzip:rewrites:ssl
       after_success: ./test/bin/k6 run test/lib/benchmark.js
       after_failure:
         - docker ps -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ jobs:
             -v $TRAVIS_BUILD_DIR/test/fixtures:/usr/local/apache2/htdocs \
             -v $TRAVIS_BUILD_DIR/test/vhosts:/usr/local/apache2/vhosts \
             -v $TRAVIS_BUILD_DIR/test/certs:/usr/local/apache2/certs \
-            -v $TRAVIS_BUILD_DIR/src:/usr/local/apache2/conf/src \
+            -v $TRAVIS_BUILD_DIR/src:/usr/local/apache2/h5bp \
             -v $TRAVIS_BUILD_DIR/httpd.conf:/usr/local/apache2/conf/httpd.conf \
             httpd
       script: ./test/bin/k6 run test/lib/index.js -e TESTS=basic-file-access:cache-busting:custom-errors:forbidden-files:enforce-gzip:precompressed-files-gzip
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,24 @@ jobs:
     - name: "Server Config Tests"
       install:
         - curl -sSL https://github.com/h5bp/server-configs-test/releases/download/1.2.2/server-configs-test.tar.gz | tar -xz -C test/
+      before_script:
+        - docker pull httpd
+        - |
+          docker run -d -p 80:80 -p 443:443 --name server \
+            -v $TRAVIS_BUILD_DIR/test/fixtures:/usr/local/apache2/htdocs \
+            -v $TRAVIS_BUILD_DIR/test/vhosts:/usr/local/apache2/vhosts \
+            -v $TRAVIS_BUILD_DIR/test/certs:/usr/local/apache2/certs \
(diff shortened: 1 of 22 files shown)


XhmikosR:
Not sure if I can review this later, but IMO we should release a 3.2.0 from the current changes, assuming we don't have breaking ones merged already.

LeoColomb:
> we should release a 3.2.0

Indeed, see #189.

> Not sure if I can review this later

What do you mean?

LeoColomb:
Updated to reduce future noise: everything required for server config is here (tests/examples/documentation).

#168 has been merged into this PR.

Only one commit alter `.htaccess` content: 88d56f1dc3648810855bed3830c1e73cc4c59aee, moving `PROTO` env variable definition where it's actually used.
Behavior is not altered. 

@XhmikosR Let me now if you can review that one, if so take your time, no stress! 😄

@Malvoz You're welcomed to review too! 🙂

LeoColomb:
Friendly bump 🙂

Malvoz:
Just a quick catch on:

> Later in another PR:
> 
>   Cleanup (remove redondant parts like #153, #185)

Remove #185 from https://github.com/h5bp/server-configs-apache/pull/188#issue-276043105, per https://github.com/h5bp/server-configs-nginx/commit/bc39e4c07d23edfc0acd3c11de88e95af5880eee

LeoColomb:
@XhmikosR Are you available for a review? Thx! 😇

XhmikosR:
Sorry, I don't have a lot of time to spend on this repo. Maybe before proceeding with merging this see if a new release should be made from the changes so far https://github.com/h5bp/server-configs-apache/compare/3.2.1...master

LeoColomb:
No problem, just let me know. 🙂 
We can release a new version before merging, for sure.

PR #191, opened 2019-05-21
Content Transformation: Update link to RFC, add mentions of Lite Mode

### LINKED ISSUE #185: `Cache-Control: no-transform` redundant w/ TLS?

Malvoz:
Changes motivated by https://github.com/h5bp/server-configs-apache/issues/185.

Closes #185 

<hr>

@LeoColomb 

https://github.com/h5bp/server-configs-apache/issues/185#issuecomment-476262268

> Side note: if we keep it, the rfc link needs to be updated to https://tools.ietf.org/html/rfc7234#section-5.2.1.6

That's the section under "Request Cache-Control Directives", we should use the section described under "Response Cache-Control Directives" which is https://tools.ietf.org/html/rfc7234#section-5.2.2.4, correct?

### DIFF
--- a/src/web_performance/content_transformation.conf
+++ b/src/web_performance/content_transformation.conf
@@ -3,11 +3,11 @@
 # ----------------------------------------------------------------------
 
 # Prevent intermediate caches or proxies (e.g.: such as the ones
-# used by mobile network providers) from modifying the website's
-# content.
+# used by mobile network providers) and browser features such as Google's
+# "Lite Mode" from modifying the website's content.
 #
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
-# https://tools.ietf.org/html/rfc2616#section-14.9.5
+# https://tools.ietf.org/html/rfc7234#section-5.2.2.4
 #
 # (!) If you are using `mod_pagespeed`, please note that setting
 # the `Cache-Control: no-transform` response header will prevent
@@ -16,6 +16,11 @@
 # to `off`, also from rewriting other resources.
 #
 # https://developers.google.com/speed/pagespeed/module/configuration#notransform
+#
+# (!) You can test the effects of content transformation applied by
+# Google's Lite Mode by visiting: https://googleweblight.com/i?u=https://www.example.com
+#
+# https://support.google.com/webmasters/answer/6211428?hl=en
 
 <IfModule mod_headers.c>
     Header merge Cache-Control "no-transform"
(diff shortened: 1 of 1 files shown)


src/web_performance/content_transformation.conf
```diff
@@ -16,6 +16,11 @@
 # to `off`, also from rewriting other resources.
 #
 # https://developers.google.com/speed/pagespeed/module/configuration#notransform
+#
+# (!) You can test the effects of content transformation applied by
+# Google's Lite Mode by visiting: https://googleweblight.com/i?u=https://www.example.com
+#
+# https://support.google.com/webmasters/answer/6211428?hl=en
```
LeoColomb: @Malvoz Can you merge this warning with the one above?
Actually a small refactoring would be awesome.

```apacheconf
# (!) Disabling content transformation may also prevent data
#     saver and other content optimization tools.
#     Notably:
#     * `mod_pagespeed`: Header will prevent `PageSpeed` from rewriting
#       HTML files, and, if the `ModPagespeedDisableRewriteOnNoTransform`
#       directive isn't set to `off`, also from rewriting other resources.
#     * Google's Lite Mode: You can test the effects of content
#       transformation applied by visiting:
#       https://googleweblight.com/i?u=https://www.example.com
#
#     https://developers.google.com/speed/pagespeed/module/configuration#notransform
#     https://support.google.com/webmasters/answer/6211428
```

Feel free to review phrasing. 🙂

Malvoz: I refactored most of it in https://github.com/h5bp/server-configs-apache/pull/191/commits/e9714082460622507813a047933c3e73a34e651f, WDYT?

Issue #194, opened 2019-07-05
Default HSTS to secure connections only - per advise in doc


Malvoz:
https://github.com/h5bp/server-configs-apache/blob/feb3a3689f83fdd4727276d6da5a184ef5caff37/src/security/strict-transport-security.conf#L44

Since [it is advised](https://github.com/h5bp/server-configs-apache/blob/feb3a3689f83fdd4727276d6da5a184ef5caff37/src/security/strict-transport-security.conf#L30) to only set this header over a secure connection, should this line not include ` env=HTTPS` (alternatively `"expr=%{HTTPS} == 'on'"`) by default? Or rather, delete this line altogether.

LeoColomb:
I think the main reason is the we prefer sending STS header on any connection over risking a never-sent situation.
That's why the two other example lines.

Malvoz:
Gotcha!

> never-sent situation

Is this explicitly due to the fact that `"expr=%{HTTPS} == 'on'"` OR `env=HTTP` historically speaking hasn't been compatible in certain  server environments for some people?

Or is this just to be safe in case Apache just "magically" does not respond with the header even though it should over HTTPS as we're telling it to?

LeoColomb:
Good question.
Actually we probably can rewrite this part: we have have removed support for versions 2.3- since its addition and `"expr=%{HTTPS} == 'on'"` should be always compatible.

PR? 😉

Malvoz:
@LeoColomb Sure, I'll get to it soon :)

PR #195, opened 2019-07-05
`Expect-CT` header (Certificate Transparency)

Malvoz:
This is a somewhat rough proposal, for [`Expect-CT`](https://httpwg.org/http-extensions/expect-ct.html). 

There's been no public discussion within this repo around Expect-CT, or if this is even needed - if it is deemed unwanted - that is completely fine. I'll perhaps need it myself anyways. :)

### DIFF
--- a/h5bp/security/expect-ct.conf
+++ b/h5bp/security/expect-ct.conf
@@ -1,0 +1,31 @@
+# ----------------------------------------------------------------------
+# | Expect-CT                                                          |
+# ----------------------------------------------------------------------
+
+# Opt-in to Certificate Transparency (CT).
+#
+# Certificate Transparency helps guard against several types of certificate-
+# based threats, including misissued certificates, maliciously acquired
+# certificates, and rogue CAs (Certificate Authorities).
+#
+# (!) When a site requires Certificate Transparency (using the `enforce`
+#     directive), they are requesting that the browser checks that any
+#     certificate for that site appears in public CT logs (1).
+#     If there are issues with certificates, the browser may display an
+#     error message to warn visitors against privacy issues.
+#
+# It is advised (2) to only serve the Expect-CT header over a secure
+# connection.
+#
+# (1) https://crt.sh/
+#     https://www.certificate-transparency.org/known-logs
+#
+# (2) https://httpwg.org/http-extensions/expect-ct.html#http-request-type
+#
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT
+# https://www.certificate-transparency.org/resources-for-site-owners
+# https://httpwg.org/http-extensions/expect-ct.html
+
+<IfModule mod_headers.c>
+    Header always set Expect-CT "max-age=86400; enforce" "expr=%{HTTPS} == 'on'"
+</IfModule>


src/security/expect-ct.conf
```diff
@@ -0,0 +1,29 @@
...
+# (1) https://www.certificate-transparency.org/known-logs
+# (2) https://httpwg.org/http-extensions/expect-ct.html#http-request-type
+#
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT
+# https://www.certificate-transparency.org/resources-for-site-owners
+# https://groups.google.com/forum/#!forum/certificate-transparency
+# https://httpwg.org/http-extensions/expect-ct.html
+
+<IfModule mod_headers.c>
+    Header always set Expect-CT "max-age=86400; enforce" "expr=%{HTTPS} == 'on'"
```
Malvoz: This line only allows the header to be set over a secure connection, this may need to change depending on resolution in https://github.com/h5bp/server-configs-apache/issues/194.

LeoColomb:
Thanks @Malvoz!
That said, I think I will postpone the merge because:
* The RFC is still [a draft](https://datatracker.ietf.org/doc/draft-ietf-httpbis-expect-ct/)
* [Browser support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT) is very low

Malvoz:
Good evening @LeoColomb :)

All good 👍

src/security/expect-ct.conf
Malvoz: Resolved by https://github.com/h5bp/server-configs-apache/issues/194#issuecomment-508881332.

src/security/expect-ct.conf
```diff
@@ -0,0 +1,28 @@
...
+# It is advised (2) to only serve the Expect-CT header over a secure
+# connection.
+#
+# (1) https://www.certificate-transparency.org/known-logs
+# (2) https://httpwg.org/http-extensions/expect-ct.html#http-request-type
+#
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT
+# https://www.certificate-transparency.org/resources-for-site-owners
+# https://groups.google.com/forum/#!forum/certificate-transparency
+# https://httpwg.org/http-extensions/expect-ct.html
```
LeoColomb: When available, this link need to be updated to the IETF RFC.

src/security/expect-ct.conf
```diff
@@ -0,0 +1,28 @@
...
+#
+# It is advised (2) to only serve the Expect-CT header over a secure
+# connection.
+#
+# (1) https://www.certificate-transparency.org/known-logs
+# (2) https://httpwg.org/http-extensions/expect-ct.html#http-request-type
+#
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT
+# https://www.certificate-transparency.org/resources-for-site-owners
+# https://groups.google.com/forum/#!forum/certificate-transparency
```
LeoColomb: Is that link really relevant? I mean we usually don't refer to forums, but only "official" documentation ressources.

src/security/expect-ct.conf
```diff
@@ -0,0 +1,28 @@
...
+# certificates, and rogue CAs (Certificate Authorities).
+#
+# (!) When a site requires Certificate Transparency (using the `enforce`
+#     directive), they are requesting that the browser check that any
+#     certificate for that site appears in public CT logs (1).
+#
+# It is advised (2) to only serve the Expect-CT header over a secure
+# connection.
+#
+# (1) https://www.certificate-transparency.org/known-logs
```
LeoColomb: \+ https://crt.sh/ ?

src/security/expect-ct.conf
Malvoz: This does exist: https://tools.ietf.org/html/draft-ietf-httpbis-expect-ct, idk why I chose to include that version. Perhaps because - and quoting [Mark Nottingham](https://github.com/mnot/) "a more readable version". But yes, I can change to the ietf draft version instead. ^^

src/security/expect-ct.conf
Malvoz: &plus; https://www.ssllabs.com/ssltest/?

src/security/expect-ct.conf
LeoColomb: Nah, I meant to the _FINAL_ version of the RFC! 👌😉

src/security/expect-ct.conf
Malvoz: https://github.com/h5bp/server-configs-apache/pull/195/commits/67d3207a3de71bc34c8d9be803e01c23fde52ee0

src/security/expect-ct.conf
LeoColomb: ? SSLlabs is not a CT logs explorer, is it?

Malvoz: It does provide information about certificate(s), including whether they're trusted or not, here's a screenshot from analyzing https://web.dev:

ssllabs

Malvoz:
> [Browser support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT) is very low

The support doesn't reach across that many browsers at this point, but Chrome has [over 50%](http://gs.statcounter.com/browser-market-share) market share according to gstats. So this is already pretty useful. ^^

src/security/expect-ct.conf
LeoColomb: https://crt.sh/ provides the complete log for a single certificate ([ex for web.dev](https://crt.sh/?id=1634127642)).
https://www.ssllabs.com/ssltest/ seems more generic, might be added in the SSL/TLS section but I'm not sure for here.

Malvoz: Resolved in https://github.com/h5bp/server-configs-apache/pull/195/commits/9dafb7bdc9fb6502910eb19403c71884e9b650d8.

src/security/expect-ct.conf
```diff
@@ -0,0 +1,31 @@
...
+# ----------------------------------------------------------------------
+
+# Opt-in to Certificate Transparency (CT).
+#
+# Certificate Transparency helps guard against several types of certificate-
+# based threats, including misissued certificates, maliciously acquired
+# certificates, and rogue CAs (Certificate Authorities).
+#
+# (!) When a site requires Certificate Transparency (using the `enforce`
+#     directive), they are requesting that the browser check that any
```
XhmikosR, line 12:
```suggestion
#     directive), they are requesting that the browser checks that any
```

Malvoz:
Closing per https://github.com/h5bp/server-configs-apache/issues/222#issuecomment-619559095.

PR #196, opened 2019-07-06
Default to HSTS only over secure connections

Malvoz:
Fix https://github.com/h5bp/server-configs-apache/issues/194.

Removes `ENV=https` example per https://github.com/h5bp/server-configs-apache/issues/194#issuecomment-508881332, and set `"expr=%{HTTPS} == 'on'"` as default.

Also replaces occurrences of "SSL" (Secure Socket Layer) with [TLS](https://tools.ietf.org/html/rfc8446) (Transport Layer Security) as [SSL is deprecated](https://tools.ietf.org/html/rfc7568).

### DIFF
--- a/src/security/strict-transport-security.conf
+++ b/src/security/strict-transport-security.conf
@@ -21,8 +21,8 @@
 #     connect over HTTP.
 #
 # (1) Preloading Strict Transport Security.
-#     To submit your site for HSTS preloading, it is required that you:
-#     * ensure the `includeSubDomains` directive is specified
+#     To submit your site for HSTS preloading, it is required that:
+#     * the `includeSubDomains` directive is specified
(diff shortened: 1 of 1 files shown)


LeoColomb:
> expr=value available in 2.4.10 and later
>
> https://httpd.apache.org/docs/current/mod/mod_headers.html#Header

Ark...

Malvoz:
Ouch.. I should've verified `expr` support first... Do you want me to just re-add the old logic here but keep small doc changes such as replacing mentions of SSL with TLS?...

XhmikosR:
I thought we already use `expr` everywhere else and the minimum Apache version is 2.4?

LeoColomb:
Indeed. OK, merging.

Issue #187, opened 2019-05-01
Ensure HTTP headers are applied to the correct Content-Type


Malvoz:
It was brought to my attention that for example, the [CSP header](https://github.com/h5bp/server-configs-apache/blob/master/src/security/content-security-policy.conf) should be sent for other resources such as XML and workers.

(While we could send any given HTTP header in all responses, it'll contribute to header bloat, and for other headers it may even cause issues)

https://github.com/w3c/webappsec/issues/520
https://github.com/webhintio/hint/issues/2342

LeoColomb:
Well, let's see the conclusions of these issues before concrete action.
I look forward to webhint answer.

Malvoz:
webhint.io settled on setting CSP for `text/html|text/xml|application/xhtml+xml|text/javascript|application/pdf|image/svg+xml` in https://github.com/webhintio/hint/pull/2618, because they're all able to execute scripts in their context (except for JS, but it's the easiest way to set [CSP for workers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#CSP_in_workers)).

So adjusting the pattern to match the types in [`media_types.conf`](https://github.com/h5bp/server-configs-apache/blob/master/src/media_types/media_types.conf): remove `xhtml+xml` and change `text/xml` to `application/xml` (both are valid but the latter matches what H5BP already does).

which equals:
`text/html|application/xml|text/javascript|application/pdf|image/svg+xml`

However `application/pdf` is currently not handled by `media_types.conf`. Should it be? IMO yes.

LeoColomb:
OK.

```regex
text\/(html|javascript)|application\/pdf|xml
```

is sufficient or are false positives for other xml relative types undesirable?

Malvoz:
Covering all cases of XML is probably even better. Yes it's sufficient IMO.

LeoColomb:
Alright. Can't wait for your PR! 😉

LeoColomb:
Closed via #200

PR #200, opened 2019-07-28
Expand responses to include CSP (per #187)

Malvoz:
Also, IMO I think there are other concerns for #187 before closing it. E.g. `Referrer-Policy`.

### DIFF
--- a/src/security/content-security-policy.conf
+++ b/src/security/content-security-policy.conf
@@ -54,6 +54,13 @@
 #      website itself to be loaded over HTTPS you must include the
 #      `Strict-Transport-Security` header.
 #
+#  (6) The `Content-Security-Policy` header is included in all responses
+#      that are able to execute scripting. This includes the commonly used
+#      file types: HTML, XML and PDF documents. Although Javascript files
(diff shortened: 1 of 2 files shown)


Malvoz:
@LeoColomb ❌ 😕 
https://travis-ci.org/h5bp/server-configs-apache/jobs/564620379

LeoColomb:
@Malvoz Yes, these failures are expected. Let me update the tests.

src/security/content-security-policy.conf
```diff
@@ -54,6 +54,13 @@
 #      website itself to be loaded over HTTPS you must include the
 #      `Strict-Transport-Security` header.
 #
+#  (6) The `Content-Security-Policy` header is included in all responses
+#      that are able to execute scripting. This includes the commonly used
+#      file types: HTML, XML and PDF documents. Although Javascript files
```
Malvoz: > Although Javascript files can not execute script in a "browsing context", they are still included to target workers

I'm not sure my terminology here is correct. I'll revisit soon.

PR #204, opened 2019-11-12
`Referrer-Policy`:  expand applicable media-types + stricter default

Malvoz:
This PR does 2 things:

- Expanding media-types for which `Referrer-Policy` should be sent, part of the effort to close out https://github.com/h5bp/server-configs-apache/issues/187.

  I've reused the same media-type matching as done in CSP, which I think is appropriate for Referrer-Policy as well. I can't say if there are more media-types to cover for sure as this isn't properly documented anywhere on the web AFAICT. Nevertheless, this should be an improvement regarding user's privacy in comparison to the current matching of `text/html` only. Some resources confirming the applicable media-types:
https://w3c.github.io/webappsec-referrer-policy/#integration-with-html  
https://w3c.github.io/webappsec-referrer-policy/#integration-with-css  
https://w3c.github.io/webappsec-referrer-policy/#integration-with-fetch  
https://github.com/w3c/webappsec-referrer-policy/issues/108
- Add a stricter default value, replacing `no-referrer-when-downgrade` with `strict-origin-when-cross-origin` for 2 reasons:

  - IMO, the example policy should've always been stricter than the default value (otherwise zero value in specifying it).
  - [Spec change](https://github.com/whatwg/fetch/pull/952) (!) to have `strict-origin-when-cross-origin` be the new default - scheduled for [release in Chrome 80](https://www.chromestatus.com/feature/6251880185331712).

I simplified the Referrer-Policy description as well, current text is more or less copy&paste from spec and is not super developer-friendly.


Malvoz:
> Spec change (!) to have strict-origin-when-cross-origin be the new default

To emphasize, if we don't change the example policy, then people who enable it as-is will downgrade and send more referrer information than the new default value.

LeoColomb:
> * scheduled for [release in Chrome 80](https://www.chromestatus.com/feature/6251880185331712).

You know my sensitivity regarding coverage. 😉 
If we merge it and deploy on servers today, is this value accepted and strong enough?

Malvoz:
Coverage is important, indeed!

To clarify, the `strict-origin-when-cross-origin` value already has quite the coverage (while there are discrepancies between [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#Browser_compatibility) and [CanIUse](https://caniuse.com/#feat=referrer-policy)). In versions of a browser where it isn't supported, it will be treated as an [unknown policy value](https://www.w3.org/TR/referrer-policy/#unknown-policy-values) and defer to the default policy (`no-referrer-when-downgrade` in current browsers, `strict-origin-when-cross-origin` in those who implement the spec change).

> strong enough?

While it is stronger than `no-referrer-when-downgrade` (see [this MDN table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#Examples)), the strongest policy is `no-referrer`, but because people love their analytics metrics, I think it's best to avoid that as the example policy (see [the notes on that](https://github.com/h5bp/server-configs-apache/pull/204/files#diff-927ed134d1452a41a2de83c2b7e76979R12-R13)).

src/security/referrer-policy.conf
```diff
@@ -2,24 +2,23 @@
...
+#     This includes the commonly used resource types:
+#     HTML, XML/SVG, PDF documents, scripts and workers.
 #
-# To check your referrer policy, you can use an online service
-# such as: https://securityheaders.io/.
+# To prevent referrer leakage entirely, specify the `no-referrer` value
+# instead. Note that the effect could impact analytics metrics negatively.
+#
+# To check your Referrer Policy, you can use an online service
+# such as: https://securityheaders.com/.
```
XhmikosR: I'd mention https://observatory.mozilla.org/ too

XhmikosR:
Do we have any info that this is indeed useful for non-HTML resources? Sorry, I don't have the time to follow this repo closely nowadays.

Malvoz:
@XhmikosR I think the most explicit information I have found for this is in the Referrer-Policy spec under the section "integration with CSS" (linked to in https://github.com/h5bp/server-configs-apache/pull/204#issue-339956409):

> ISSUE 1&nbsp;&nbsp;&nbsp;This requires that CSS style sheets process `Referrer-Policy` headers, and store a referrer policy in the same way that Documents do.

I interpret that particular section as advice for developers to set the header for stylesheets.

But I thought I'd provide a test (includes test for CSS, JS and SVG) to confirm - and indeed those resource would send `Referer` for requests even if the HTML document sent `Referrer-Policy: no-referrer`.


Open `page.html` in [referrer-policy.zip](https://github.com/h5bp/server-configs-apache/files/3853727/referrer-policy.zip) and also see the comments in `.htaccess` that also explains it (while I realize you might not have time, perhaps @LeoColomb can confirm this for you).

src/security/referrer-policy.conf
Malvoz: Added.

Malvoz:
And partially confirmed in https://www.chromestatus.com/feature/5696724538490880.

PR #223, opened 2020-04-13
Release v4.0.0

LeoColomb:


XhmikosR:
A couple of comments:

1. should CI pass?
2. shouldn't we include the new folders/files in the npm package? Currently only the dist file is included https://github.com/h5bp/server-configs-apache/blob/6bf8df6f5977c372ce81c91b1de5fa76321a7b44/package.json#L22

LeoColomb:
> 1. should CI pass?

They are all passing 😆 ([service was degraded](https://www.githubstatus.com/incidents/2k3z9fkr9z5g?u=5m5p998sn033) right when this PR was submitted)

> 2\. shouldn't we include the new folders/files

NPM is only for `.htaccess` distribution, I don't want to change this, and definitely not for the whole server configuration.
Let me know what do you think.

XhmikosR:
> 
> 
> > 1. should CI pass?
> 
> They are all passing 😆 ([service was degraded](https://www.githubstatus.com/incidents/2k3z9fkr9z5g?u=5m5p998sn033) right when this PR was submitted)
> 
> > 2. shouldn't we include the new folders/files
> 
> NPM is only for `.htaccess` distribution, I don't want to change this, and definitely not for the whole server configuration.
> Let me know what do you think.

Your call, I was basically just pointing this out in case it was a mistake.

LeoColomb:
All fine, thanks for this! 👍

LeoColomb:
Not so far from the perfection now 😉 Thanks a lot again!
Ready for the release? 🙂

XhmikosR:
Yup, go ahead :shipit:

XhmikosR:
Ah, it seems the dist file is not up to date. Not a big deal, just a FYI.

LeoColomb:
Ah good one @XhmikosR, I'll rebuild it in a minute.
At least thanks to my incredible work on GitHub Actions workflows, npm version is fine 😉

Issue #234, opened 2020-05-07
Centos8 requires DSO from conf.modules.d


zishanj:
Centos8 with Apache v2.4.x requires loading DSO from `conf.modules.d` directory otherwise HTTPD service always fails to start without leaving any meaningful error in logs except graceful shutdown of the service. In `httpd.conf` file, its required to include this line (as mentioned in original Apache httpd.conf file for Centos):

> Include conf.modules.d/*.conf

This should be mentioned somewhere in Readme file too for ease of others.

LeoColomb:
Thanks for opening this issue @zishanj.

Do you have the appropriate documentation for this requirement?

zishanj:
This is what mentioned in original httpd.conf file that comes with Centos version of Apache:

> #
> # Dynamic Shared Object (DSO) Support
> #
> # To be able to use the functionality of a module which was built as a DSO you
> # have to place corresponding `LoadModule' lines at this location so the
> # directives contained in it are actually available _before_ they are used.
> # Statically compiled modules (those listed by `httpd -l') do not need
> # to be loaded here.
> #
> # Example:
> # LoadModule foo_module modules/mod_foo.so
> #
> Include conf.modules.d/*.conf

This `conf.modules.d` directory on Centos version of Apache includes `00-systemd.conf` which loads the `systemd` module:

`LoadModule systemd_module modules/mod_systemd.so`

This is **required** on Centos to get feedback while activating the service otherwise it will shutdown itself _gracefully_ with error in log like:

> caught SIGWINCH, shutting down gracefully

zishanj:
Here is complete list of other files that are loaded with `Include conf.modules.d/*.conf`

![image](https://user-images.githubusercontent.com/1451292/81379644-e8ddff80-9122-11ea-8aad-a603ed5808d7.png)

LeoColomb:
Thanks for this @zishanj.

Could you past here the full content of `00-systemd.conf`?
Thanks again.

zishanj:
There is just a `LoadModule` line which I mentioned before:

`LoadModule systemd_module modules/mod_systemd.so`

LeoColomb:
Ah, that's nice, thanks @zishanj!

Finally, do we agree everything works fine on CentOS if we add that single `LoadModule` line at the end of the `LoadModule` block on the `httpd.conf`?

[...]

PR #235, opened 2020-05-10
Added systemd module to support CentOS

zishanj:
CentOS requires `systemd` module to interact with `httpd` service otherwise it will shutdown gracefully leaving error in log:

> caught SIGWINCH, shutting down gracefully


zishanj:
#234

PR #242, opened 2020-07-31
apache error notification

clepces:
Perform a small test with the templates in vhosts / templates / example.com.conf and the file .000-default.conf and apache and if I get a result of notification of denial of loading.
read the documentation and it is only loaded in httpd and outside the virtualhost tags.
(x_x) I'm new to the subject and I'm learning.

### DIFF
--- a/h5bp/ssl/ssl_engine.conf
+++ b/h5bp/ssl/ssl_engine.conf
@@ -15,6 +15,9 @@
 #     By enabling a cache, we tell the client to re-use the already
 #     negotiated state.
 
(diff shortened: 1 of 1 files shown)


LeoColomb:
Thanks for opening this PR, @clepces!

> read the documentation and it is only loaded in httpd and outside the virtualhost tags.

Can you provide the link to this documentation? Thx!

LeoColomb:
I'm closing this PR as inactive.
If you see still see the behavior described, don't hesitate to continue the discussion below.

PR #255, opened 2020-11-26
Add php expiration for no cache

lucalbert:
Hi,

If PHP is no longer available, the server sends the raw file with `Content-Type: application/x-httpd-php` and uses the default expiration value of 1 month for this content type.

This causes serious problems because when the PHP pool goes up, the browser continues to serve this raw file.

This is fixed by advising not to cache the `application/x-httpd-php` type.

### DIFF
--- a/dist/.htaccess
+++ b/dist/.htaccess
@@ -1102,6 +1102,11 @@ FileETag None
     ExpiresByType text/html                             "access plus 0 seconds"
 
 
+  # PHP
+
+    ExpiresByType application/x-httpd-php               "access plus 0 seconds"
+
+
   # JavaScript
 
     ExpiresByType application/javascript                "access plus 1 year"
--- a/h5bp/web_performance/cache_expiration.conf
+++ b/h5bp/web_performance/cache_expiration.conf
@@ -45,6 +45,9 @@
 
     ExpiresByType text/html                             "access plus 0 seconds"
 
+  # PHP
+
+    ExpiresByType application/x-httpd-php               "access plus 0 seconds"
 
   # JavaScript
 


LeoColomb:
Thanks for opening this PR @lucalbert.
Is there some documentation available online that cover this behavior?

lucalbert:
Hello @LeoColomb,

I am not sure that there is a precise documentation on this point (at the level of the expiration in any case) because it is clearly a problem either of configuration of the vhost (PHP incorrectly configured, or not- started), or a technical problem that prevents PHP from working properly.

There are quite a few topics including this one: https://serverfault.com/questions/785782/apache-with-php-fpm-php-doesnt-execute

But I think the type `application/x-httpd-php` should behave the same as` text/html` i.e. 0 seconds and not the default of 1 month?

LeoColomb:
Thanks for your reply.
The link provided does not mention `application/x-httpd-php` usage 😉 
I have few concerns about the following:
* This MIME-type is very specific to PHP, we try to keep h5bp config free of server language specific attribute.
* The same MIME-type is not a standard one, so might be better to configure at the application level and not at the server level.

LeoColomb:
@lucalbert I tend to agree with @quantumpacket.

> > If PHP is no longer available, the server sends the raw file with Content-Type: application/x-httpd-php and uses the default expiration value of 1 month for this content type.
> 
> @lucalbert a properly configured server should **NEVER** send the raw PHP file to a client, even if the pool goes down. This is a huge security risk. Fixing your security issue by stopping caching of the raw file does nothing to fix your insecure configuration. The same applies to [h5bp/server-configs-apache#255](https://github.com/h5bp/server-configs-apache/pull/255)

https://github.com/h5bp/server-configs-nginx/pull/263#issuecomment-734472025

What do you think about this @lucalbert?

LeoColomb:
@lucalbert I'm closing the PR for now, as per previous comments.
Don't hesitate to continue the discussion below.

PR #271, opened 2021-06-14
Refresh security policies headers

### LINKED ISSUE #179: `Permissions-Policy` header
When scanning my site through https://securityheaders.com/ I got flagged because I was missing a `Feature-Policy` header: https://scotthelme.co.uk/a-new-security-header-feature-policy/

Would you be interested in adding it to the template? I can make a PR if that's the case.

---

* Paper: https://w3c.github.io/webappsec-permissions-policy/
* MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy
* CanIUse: https://caniuse.com/permissions-policy

### LINKED ISSUE #190: `Content-Security-Policy` disallow 'object-src' by default
Google's [csp-evaluator](https://csp-evaluator.withgoogle.com/) is complaining whenever the fetch-directive `object-src` is missing (unless `default-src` is `'none'`, which is rarely done in the wild). I [reported this to webhint](https://github.com/webhintio/hint/issues/25#issuecomment-488153695) aswell.

`object-src`, per the spec, "acts upon any request made on behalf of an `<object>`, `<embed>`, or `<applet>` element" and can be used to allow [`plugin-types`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/plugin-types) such as `application/x-shockwave-flash` (also, this directive [may be removed](https://github.com/w3c/webappsec-csp/issues/394) from the web platform).

These elements are largely considered legacy elements that do not recieve new standardized security features such as (`<iframe>`) `sandbox` or `allow` attributes.

The recommendation is documented in https://csp.withgoogle.com/docs/strict-csp.html, and https://csp.withgoogle.com/docs/faq.html#comparison:

> Since setting directives other than `script-src` and `object-src` doesn't increase the protection against cross-site scripting, and it adds adoption and maintenance costs of CSP, we believe most applications should focus on deploying the baseline strict policy as the most high-impact improvement.

I don't have knowledge of how common it is or which other types of plugins may be used. I think people use `<object>` to embed PDFs etc. However, I don't think developers implementing a CSP today is still using these legacy elements.

WDYT?

### LINKED ISSUE #192: `Access-Control-Allow-Origin` pointing to an origin should include `Vary: Origin`
When `Access-Control-Allow-Origin` points to a URL rather than `*`, then `Vary: Origin` should be sent along with the response.

Sources:
- https://fetch.spec.whatwg.org/#cors-protocol-and-http-caches
- https://w3c.github.io/webappsec-cors-for-developers/#use-vary
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#CORS_and_caching
- https://textslashplain.com/2018/08/02/cors-and-vary/

All of the [CORS configs](https://github.com/h5bp/server-configs-apache/tree/master/src/cross-origin) use `*` in the examples. But if anything, there could potentially be a note of this in [`requests.conf`](https://github.com/h5bp/server-configs-apache/blob/master/src/cross-origin/requests.conf)? Or something better than just a note?

### LINKED ISSUE #250: Add `Cross-Origin-Embedder-Policy` and `Cross-Origin-Opener-Policy`
Please can you add `Cross-Origin-Embedder-Policy` and `Cross-Origin-Opener-Policy` to the `/h5bp/security/` section.

More info: https://web.dev/coop-coep/

More info: https://developers.google.com/web/updates/2020/10/devtools#frame-details

As we want to add it to the apache header list.

LeoColomb:
* [x] Add Cross Origin Policies headers
  Fix #250
* [x] Add `Permissions-Policy` header
  Fix #179
* [x] Tweak current header directives
  Fix #192
  Fix #190 
* [x] Validate relevant mime-types & documentation
* [x] Reflect on `.htaccess`


LeoColomb:
@Malvoz You're welcome to review this one, as it covers some of your reports. 🙂

PR #286, opened 2021-07-27
Adding django server deployment template

Tikam02:
- Added Django WSGI configuration on templates
- commented Include h5bp/tls/policy_intermediate.conf, because it didn't exists on /h5bp/tls/  on latest commit.

Added this because people couldn't find precise configuration files while deploying dev or production server, this repo is perfect and making standard, so I thought let's contribute more templates to it.

Please Review this configuration, so that if any issue I will work on that and make PR again.
Thanks !!


LeoColomb:
Thanks for this pull request @Tikam02.
That said the suggestion is too project specific to be included.

PR #287, opened 2021-08-02
Remove charset from json and webmanifest

Rotzbua:
The charset is not required and useless according to the IANA:

> Note:  No "charset" parameter is defined for this registration.
>   Adding one really has no effect on compliant recipients.

https://www.iana.org/assignments/media-types/application/json
https://www.iana.org/assignments/media-types/application/manifest+json

### DIFF
--- a/h5bp/media_types/character_encodings.conf
+++ b/h5bp/media_types/character_encodings.conf
@@ -23,7 +23,6 @@ AddDefaultCharset utf-8
                      .htc \
                      .ics \
                      .js \
-                     .json \
                      .manifest \
                      .map \
                      .markdown \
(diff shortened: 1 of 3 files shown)


mathiasbynens:
Reality does not always match IANA, unfortunately. Have you tested this in practice? E.g. with this change, does `https://example.com/foo.json` render non-ASCII characters correctly when viewed directly in a browser?

I’d personally rather not land this.

LeoColomb:
Thanks for opening this pull request @Rotzbua.
This is true in theory, that said I second @mathiasbynens for the practice.
Additionally:
* Charset and mime types are enforced as per https://github.com/jshttp/mime-db values. Your concern should be discussed there instead.
* Actually already discussed there: https://github.com/jshttp/mime-db/pull/225

PR #295, opened 2021-11-24
Add realpath to build.sh

### LINKED ISSUE #294: Correctly resolve path of build script
The **bin/build.sh** script does (at least on MacOS) not correctly resolve it's file path when the build.sh script has been symlinked: Variable "$0" in build.sh will not contain the 'real' file name but the symlinks path rather, e.g. `/Users/me/bin/build-htaccess` instead of expetcted `/Users/me/server-configs-apache/bin/build.sh`

The following **$repo_root** determination is then based on false assumptions. Adding `realpath` could fix this:

```bash
# repo_root=$(dirname "$(dirname "$0")")
repo_root=$(dirname "$(dirname "$(realpath "$0")")")
```

What do you think?

tomkyle:
Correctly resolve build.sh's directory name even when script has been symlinked. Fixes #294

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -3,7 +3,7 @@
 declare htaccess_config_default="bin/htaccess.conf";
 declare htaccess_output_default="./.htaccess"
 declare repo_root
-repo_root=$(dirname "$(dirname "$0")")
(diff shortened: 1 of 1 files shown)


LeoColomb (approved):
Great, thanks @tomkyle

PR #318, opened 2022-09-19
Expand ~ in filename

### LINKED ISSUE #317: Variable evaluation in build script
Hey!

I'm currently facing an issue with the build script. I'm trying to generate a custom htaccess file while including a custom partial in it. Here's a basic example:

```apache
title   "cross-origin"
enable  "h5bp/cross-origin/images.conf"
enable  "h5bp/cross-origin/web_fonts.conf"
enable "~/my/custom/path/custom-conf.conf"
```

Problem is `~` is never evaluated, neither `$HOME`. 
https://github.com/h5bp/server-configs-apache/blob/06df209e2e24832124e09f4f88d4bf4cf1cf4d38/bin/build.sh#L52

I suck in bash, do you think it would possible to enable this?

nlemoine:
Fixes #317

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -37,6 +37,7 @@ create_htaccess() {
         # Remove quotes surrounding
         filename="${filename%\"}"
         filename="${filename#\"}"
+        filename="${filename/#\~/$HOME}"
 
         # Evaluate
         case "${keyword}" in


LeoColomb:
Pull request closed as per comment on #317.

Issue #332, opened 2023-01-08
Remove `<IfModule>` from `Options -Indexes`


bchr:
Hey. You have this snippet in your code:

https://github.com/h5bp/server-configs-apache/blob/14d8af4ac96a0918713f63e370f894b4305bb025/h5bp/security/file_access.conf#L11-L13

`Options` is a [part of `core`](https://httpd.apache.org/docs/2.4/mod/core.html#options), not `autoindex`.

Another place in your code you have `Options -MultiViews` which is correctly not surrounded by `<IfModule>`.

My suggestion is to remove the surrounding `<IfModule>` from `Options -Indexes`.

Eventually see https://htaccessbook.com/disable-directory-indexes/

Thanks!

LeoColomb:
Thanks for opening this issue, @bchr.
Indeed, you are right, the surrounding `<IfModule>` can be removed.
Would you like to send a pull request?

PR #335, opened 2023-01-10
Remove `<IfModule>` block for `Options -Indexes`

bchr:
Fixes https://github.com/h5bp/server-configs-apache/issues/332


LeoColomb (approved):
Thanks @bchr!

PR #341, opened 2023-02-07
Document pre-compressed content usage in Virtual Hosts

egfx-notifications:
I used the rules for delivering precompressed brotli and gzip content for my server, but neither using `RewriteOptions Inherit` nor pasting the rules directly into the Virtual Host config enabled my Virtual Hosts to actually serve the precompressed content.

I activated tracing logs for mod_rewrite and saw that `REQUEST_FILENAME` contained the path of the request instead of a final filesystem path. Reading up on mod_rewrite documentation I found out that this can indeed happen depending on the current state of the URI to file translation of the request. Virtual Hosts are one case where the variable usually still contains the request URI instead of a filesystem path. [The documentation mentions a lookahead, though](https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html#rewritecond:~:text=If%20used%20in,value%20of%20REQUEST_FILENAME.). This PR implements this lookahead so the rule also works in contexts such as Virtual Hosts.

### DIFF
--- a/h5bp/web_performance/pre-compressed_content_brotli.conf
+++ b/h5bp/web_performance/pre-compressed_content_brotli.conf
@@ -13,14 +13,11 @@
 #     default resources priorities.
 #     https://httpd.apache.org/docs/current/mod/mod_dir.html#directoryindex
 #
-# (1) REQUEST_FILENAME will only contain a filepath if the translation
-#     from URI to filepath has already been completed.
-#     If the translation is still pending the variable is equal to REQUEST_URI
-#     Translation is usually still pending when using it in VirtualHost context.
-#     We can use a lookahead by prepending LA-U:
-#     This way we can check the final filepath even in contexts where it
-#     would not be possible otherwise.
-#     https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html#rewritecond 
+# (1) In a virtual host context, `REQUEST_FILENAME` may be equal to
+#     `REQUEST_URI` while URL translation to file path is still pending.
+#     In that case, enabling the look-ahead flag on `RewriteCond` (`LA-U`)
+#     will extract URI last segment to work around a file path resolution.
+#     https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewritecond 
 #
 # (2) Remove default Content-Language header added for .br files.
 #     https://httpd.apache.org/docs/current/mod/mod_mime.html#multipleext
@@ -35,7 +32,8 @@
 
     RewriteCond %{HTTP:Accept-Encoding} br
     # (1)
-    RewriteCond %{LA-U:REQUEST_FILENAME}\.br -f
+    RewriteCond %{REQUEST_FILENAME}\.br -f
+    # RewriteCond %{LA-U:REQUEST_FILENAME}\.br -f
     RewriteRule \.(css|ics|js|json|html|svg)$ %{REQUEST_URI}.br [L]
 
     # Prevent mod_deflate double gzip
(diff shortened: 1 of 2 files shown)


LeoColomb:
Thanks for submitting this pull request, @egfx-notifications.
Could you elaborate your use case where `REQUEST_URI` (default value of `REQUEST_FILENAME`) is an issue?

egfx-notifications:
The conditions `RewriteCond %{REQUEST_FILENAME}\.br -f` and `RewriteCond %{REQUEST_FILENAME}\.gz -f` in the precompressed content rules never match, as a request to e.g `example.com/assets/style.css` would result in `REQUEST_URI` `/assets/style.css`, this is interpreted as absolute filesystem path and thus never found. So precompressed content is never served.  
This can be fixed by either prepending `DOCUMENT_ROOT` in which case it only works for non-resolved paths, or with the lookahead in this PR which always resolves to the final filesystem path regardless of processing stage.

Note this is not an issue if used in `.htaccess` or Directory sections as the URI to path translation has already been done by then. It is only an issue when using this prior to translation, e.g. in VirtualHost sections. I have the rules in my default server config and import them into the Virtual Hosts with `RewriteOptions Inherit`, thus the current solution without lookahead does not work.

LeoColomb:
Alright thanks, crystal clear. 👍
I'm quite concerned by the failing tests, that said.

LeoColomb:
Which make me guess that you have another level of rewrite, right?
As usually the file path matches

egfx-notifications:
Ok, I tested this locally, it does not work with the test cases because those are included in a `<Directory>` section, thus `REQUEST_FILENAME` already contains the resolved filepath instead of the same value as `REQUEST_URI`. In contrast to my assumptions the lookahead does not happen only for unresolved URIs, but the already resolved path is then treated as a new `REQUEST_URI` for the internal subrequest which the lookahead makes, ultimately ending in a too many recursions error once the limit for internal subrequests is reached, oops.

So to make this work in all scenarios we would have to use the existing solution in `<Directory>` sections and .htaccess, but the lookahead for `<VirtualHost>` sections and possibly others that are not resolved yet.

I tried to find a stable way to implement this and this kind of works
```
    RewriteCond %{HTTP:Accept-Encoding} gzip
    RewriteCond %{REQUEST_FILENAME}\.gz -f
    RewriteRule \.(css|ics|js|json|html|svg)$ %{REQUEST_URI}.gz [L]

    RewriteCond %{HTTP:Accept-Encoding} gzip
    RewriteCond expr "%{REQUEST_FILENAME} == %{REQUEST_URI}"
    RewriteCond %{LA-U:REQUEST_FILENAME}\.gz -f
    RewriteRule \.(css|ics|js|json|html|svg)$ %{REQUEST_URI}.gz [L]
```

I do not like the duplication and I do not think this will work if any rewrites happen from the original `REQUEST_URI` to another URI. Can you think of a reliable way to execute the correct `RewriteCond` in all scenarios?

Of course we could just include a note about having to use the lookahead in Virtual Host sections in the comments and leave it up to the user to select the correct `RewriteCond` for their individual deployment.

LeoColomb:
Thank you for your extensive investigation, appreciated.

> Of course we could just include a note about having to use the lookahead in Virtual Host sections in the comments and leave it up to the user to select the correct `RewriteCond` for their individual deployment.

This looks to be the most reasonable. It is quite unfortunate to exclude such a use case per default, but I agree this duplication would be too much.

This means you can keep your PR changes commented out and add the original directive back.

LeoColomb (approved):

Issue #329, opened 2022-12-18
Add build parameter to generate `.htaccess` without comments


pauloxnet:
Currently, two-thirds of the file generated by the build command is comments. It might be useful to be able to generate a file without comments to be lighter and faster to read, leaving the option to continue generating the version with comments. I imagine an attribute to be passed to the script might be fine for this purpose.

LeoColomb:
Thanks for opening this issue @pauloxnet.
While this is a great idea, the implementation in the script will be quite challenging, if even reasonably possible.
The comments are already written in the files from which the script is generating the `.htaccess`, and there isn't any content processing so far.

If you want to remove all the comments in the generated file, I'd recommend running the following simple yet efficient command (might not be the most optimized, though). 
```shell
sed -i '/^#/d' .htaccess
```

pauloxnet:
> Thanks for opening this issue @pauloxnet.
> While this is a great idea, the implementation in the script will be quite challenging, if even reasonably possible.
> The comments are already written in the files from which the script is generating the `.htaccess`, and there isn't any content processing so far.
> 
> If you want to remove all the comments in the generated file, I'd recommend running the following simple yet efficient command (might not be the most optimized, though). 
> ```shell
> sed -i '/^#/d' .htaccess
> ```

Thanks for your answer 😃

PR #342, opened 2023-03-19
Control over comments included in `.htaccess` build

edschindler:
With reference to [issue 329](https://github.com/h5bp/server-configs-apache/issues/329) raised by @pauloxnet, which seemed useful to me as well, particularly for production deployments.

The new directive "comments" optionally filters out comments as the modules are being read in, and this reduces the complexity of the build script changes quite a bit, but does not result in 100% removal. (To achieve more complete removal, one would have to change all "Disabled" directives to "Omit", and remove all Title directives, in addition to using the new "comments none" directive - and even then, the build header remains.) Still, it provides good flexibility for builders.

The directive distinguishes two levels of comment: normal and what I'm calling "guidance", which are designated by ## rather than #. It is possible to filter out the guidance lines and leave the normal comments. This instrumentation has been applied to all the module files, along with some blank line rationalization to avoid excess vertical white space.

Existing htaccess.conf files will work without modification, as will any custom module files.

The default htaccess.conf creates an .htaccess file of 56.5K. Using "comments no-guidance", this is reduced to 29.7K. Using "comments none", it goes down to 17.2K. Changing the Disableds to Omits brings it down to 10.6K, and removing the Titles to 9.3K.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,25 +303,20 @@ enable "h5bp/rewrites/rewrite_http_to_https.conf"
 
 ##### Controlling the size of the output .htaccess file
 
-The default modules contain significant number of comment lines, which
-contain valuable guidance about how and why to use the setting, as well
+The default partials contain significant numbers of comment lines, which
+contain valuable guidance about how and why to use the setting as well
 as web references. However, some may feel that the size of the resulting
 .htaccess file is too large to deploy in a production environment.
 
-The special directive "comments" can be used to control which comment
-lines are copied out of the source modules into .htaccess. Options
-are "all" (the default), "no-guidance", and "none". Note that this
-directive does not control comments created by the build script (such
-as Title or Disabled modules), only those that are in the source
-module files.
+The special keyword "no-partials-comments" can be used to prevent
+comment lines from being copied out of the partials into .htaccess.
+Note that this keyword does not control comments created by the build
+script (such as Title or Disabled modules), only those that are in the
+source partials files.
 
 ```apache
-comments "no-guidance"
+no-partials-comments
 ```
-
-Note: within the module files, guidance comments are denoted by
-a double hash (##).
-
 
 ##### Adding custom modules
 
--- a/README.md
+++ b/README.md
@@ -306,13 +306,14 @@ enable "h5bp/rewrites/rewrite_http_to_https.conf"
 The default partials contain significant numbers of comment lines, which
 contain valuable guidance about how and why to use the setting as well
 as web references. However, some may feel that the size of the resulting
-.htaccess file is too large to deploy in a production environment.
+`.htaccess` file is too large.
 
-The special keyword "no-partials-comments" can be used to prevent
-comment lines from being copied out of the partials into .htaccess.
-Note that this keyword does not control comments created by the build
-script (such as Title or Disabled modules), only those that are in the
-source partials files.
+The special keyword `no-partials-comments` can be used to prevent
+comment lines from being copied out of the partials into `.htaccess`.
+
+> **Note** This keyword does not control comments created by the build
+> script (such as `title` or `disabled` directives), only those that
+> are in the source partials files.
 
 ```apache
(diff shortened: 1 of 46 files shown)


LeoColomb (changes requested):
Thanks for submitting this PR, @edschindler.
That being said, the number of changes does not seem reasonable for what it aims to achieve. This was the concern raised on #329.

Your submission also contains the following that cannot be merged:
* All EOF lines are removed;
* The comment structure is changed (current structure is standardized across H5BP repositories).

To keep it short: while I'm not against the goal, I can't accept this PR as is. Except if you have a lighter alternative, it's going to be rejected.

edschindler:
Thanks for your review, @LeoColomb .

Of course the changes to the individual modules are not required, I saw them as relatively inconsequential (even though they touched a lot of files) once the build capability had been added. They can all be rolled back, and the "no-guidance" feature (which uses ##) removed.

Should I interpret the change that you've made to the title as a steer towards a command line parameter when running the shell script (rather than a directive inside the .conf file)? Points:

1) We don't have a full command line parser active, and adding the parameter as a command line option requires providing both of the other parameters (.conf file location and .htaccess destination) before adding the build parameter to the command line in the third position. I think this hurts usability, but on the plus side it would not require a .conf edit to build different versions of the same configuration.

2) The directive as it stands is not really a "build parameter" since its state can be changed mid-build if desired. If it remains inside the .conf file, it would add more complexity to try to limit it to being applied only one time, to the entire build.

Given all of that, your initial reaction to issue 329 was that it would be too complicated to achieve. That issue, and your reaction, did not anticipate the wholesale changes that I made to the modules. So aside from these module changes, did I generally succeed in keeping the complexity of the solution (the directive, and the two-stage filtering) at an acceptable level? Because I think the value is lost with still simpler approaches (one might as well just strip the comments after the build, as you suggested) and I am happy to step away now if so.

edschindler:
Gave it a try...

README.md
```diff
@@ -301,6 +301,23 @@ snippet, change the `disable` keyword to `enable`:
 enable "h5bp/rewrites/rewrite_http_to_https.conf"
 ```
 
+##### Controlling the size of the output .htaccess file
+
+The default partials contain significant numbers of comment lines, which
+contain valuable guidance about how and why to use the setting as well
+as web references. However, some may feel that the size of the resulting
+.htaccess file is too large to deploy in a production environment.
```
LeoColomb, line 309:
```suggestion
`.htaccess` file is too large.
```

README.md
```diff
@@ -301,6 +301,23 @@ snippet, change the `disable` keyword to `enable`:
...
+
+The special keyword "no-partials-comments" can be used to prevent
```
LeoColomb, line 311:
```suggestion
The special keyword `no-partials-comments` can be used to prevent
```

README.md
```diff
@@ -301,6 +301,23 @@ snippet, change the `disable` keyword to `enable`:
...
+comment lines from being copied out of the partials into .htaccess.
```
LeoColomb, line 312:
```suggestion
comment lines from being copied out of the partials into `.htaccess`.
```

README.md
```diff
@@ -301,6 +301,23 @@ snippet, change the `disable` keyword to `enable`:
...
+Note that this keyword does not control comments created by the build
+script (such as Title or Disabled modules), only those that are in the
+source partials files.
```
LeoColomb, line 315:
```suggestion

> **Note** This keyword does not control comments created by the build
> script (such as `title` or `disabled` directives), only those that
> are in the source partials files.
```

LeoColomb (changes requested):
Indeed, much more reasonable, and actually, it looks pretty good.
Thanks, @edschindler!
Few documentation tweaks.

edschindler:
Happy to have been able to help @LeoColomb , and thanks for your guidance...

Issue #359, opened 2024-03-01
Format `Content-Security-Policy` header directives


dcog989:
After implementing the included CSP and running it against e.g. https://csp-evaluator.withgoogle.com/ and others, plus a lot of reading, I came up with:

```apacheconf
<IfModule mod_headers.c>
    Header always set Content-Security-Policy "
        default-src 'self' gap:;
        style-src 'self' https://fonts.googleapis.com;
        font-src 'self' data: https://fonts.gstatic.com;
        img-src 'self' data: content:;
        script-src 'nonce-rAnd0m' 'strict-dynamic';
        media-src 'self';
        base-uri 'none';
        form-action 'self';
        frame-ancestors 'none';
        object-src 'none';
        require-trusted-types-for 'script';
        upgrade-insecure-requests"
        "expr=%{CONTENT_TYPE} =~ m#text\/(html|javascript)|application\/pdf|xml#i"
</IfModule>
```

I think that provides a better launch pad for users than the current - both formatting for readability and a more complete list of arguments.

LeoColomb:
Thanks for opening this issue @dcog989.
The one-per-line-attribute rule sounds good, would you like to open a pull request?
On their actual values, let's keep them as strict per default for now.

dcog989:
Hey. Yeah, I'd be happy to do that.

I was also wondering if the project would be interested in a significant reformat of the entire document, which I think could be improved for brevity and clarity. E.g.:

Existing:

```"Using `.htaccess` files slows down Apache, therefore, if you have access to the main server configuration file (which is usually called `httpd.conf`), you should add this logic there."```

Proposed:

```".htaccess slows down Apache so use the main server configuration file (usually `httpd.conf`) if possible."```

Also, improve consistency of comments and section formatting using syntax from https://github.com/aaron-bond/better-comments (even if a user doesn't employ that extension, the comments would have useful notation?). E.g.:

![Screenshot 2024-03-11 212403](https://github.com/h5bp/server-configs-apache/assets/89043002/65dbab5d-ed51-4f53-a3f6-89c9ade1fa9f)

LeoColomb:
> Hey. Yeah, I'd be happy to do that.

Cool, thanks! 👍

> I was also wondering if the project would be interested in a significant reformat of the entire document, which I think could be improved for brevity and clarity. E.g.:

Let's stay simple for now, one thing at a time.
Also, `.htaccess` file is auto-generated, based on the files already in a proper structure.

[...]

PR #361, opened 2024-03-12
Format `Content-Security-Policy` header directives

dcog989:
Format Content-Security-Policy header directives #359


dcog989:
Ah! I understand now! Will adjust. :)

PR #362, opened 2024-03-12
Format Content-Security-Policy header directives #359

dcog989:


LeoColomb (changes requested):
Thanks for your PR @dcog989.
As you can read in the checks of the PR, the current config is not valid.
Have you tested it on your side?
I [believe](https://serverfault.com/questions/752176/break-up-a-long-line-in-a-htaccess-file) the following rules apply for multiline expression:

- The last character on all lines except the final one must be a backslash;
- The final line must not terminate with a backslash;
- The Apache comment character (#) cannot be used to comment out a line.
- You cannot break the flags

dcog989:
Hi Léo. I saw the errors but the meaning wasn't clear to me. I'm new to github and not familiar with all the tools available - and can't find a way to perform the test done when pull request submitted.

However, I think the reason for failure was the location of comment line:

```
<IfModule mod_headers.c>
    Header always set Content-Security-Policy "
	# (1) (2) (3) (4) (5) (6) (7)```

should be:

```
<IfModule mod_headers.c>
	# (1) (2) (3) (4) (5) (6) (7)
    Header always set Content-Security-Policy "```

I'll attempt to update / merge....

LeoColomb (changes requested):
Same review I'm afraid, the current config is not valid.
Have you tested it on your side?
I [believe](https://serverfault.com/questions/752176/break-up-a-long-line-in-a-htaccess-file) the following rules apply for multiline expression:

- The last character on all lines except the final one must be a backslash;
- The final line must not terminate with a backslash;
- The Apache comment character (#) cannot be used to comment out a line.
- You cannot break the flags

---
Test output can be seen in here: https://github.com/h5bp/server-configs-apache/actions/runs/8256582544/job/22646947712?pr=362#step:4:778

LeoColomb (approved):
It looks all right now, thanks @dcog989 😊

PR #371, opened 2024-04-18
Add explanatory text in case `ExpiresDefault` is not applied by default

creopard:
Added explanatory text in case "ExpiresDefault" is not applied by default.


h5bp/web_performance/cache_expiration.conf
```diff
@@ -7,6 +7,13 @@
 # (!) If you don't control versioning with filename-based cache busting, you
 #     should consider lowering the cache times to something like one week.
 #
+# (!) When using `.htaccess` file, the webserver config may have already
+#     a preset some file types.
```
bchr, line 11: Typo?

PR #394, opened 2026-07-13
Document rewrite inheritance in virtual host templates

### LINKED ISSUE #337: Add note about RewriteOptions Inherit ?
If you use the h5bp Apache config in a server configuration, and add a virtual host and a directory block with custom rewrite rules, none of the h5bp rewrite rules will work. This is because the new rewrite rules overwrite existing ones by default.

Do you think it's worth mentioning this anywhere?

E.g.

Let's say you have this (taken from h5bp) in your main config file:

```
<IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteCond %{REQUEST_URI} "!(^|/)\.well-known/([^./]+./?)+$" [NC]
    RewriteCond %{SCRIPT_FILENAME} -d [OR]
    RewriteCond %{SCRIPT_FILENAME} -f
    RewriteRule "(^|/)\." - [F]
</IfModule>
```

You then create a virtual host configuration:

```
<VirtualHost *:80>
    ServerName example.com
   # ...
   <Directory /var/www/httpdocs/>
        RewriteEngine On
        # Without inheritance of rewrite rules, none of the h5bp rules take effect :(
        RewriteCond %{REQUEST_FILENAME} !-f
        RewriteCond %{REQUEST_FILENAME} !-d
        RewriteRule . /index.php [L]
   </Directory>
</VirtualHost>
```

For blocking access to hidden files I prefer to use `Require all denied` with a `LocationMatch`, but this is just an example.

It strikes me that it's not immediately obvious that the rules from h5bp have no effect in certain situations. At least, it wasn't to me.

Ref: https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriteoptions

oiahoon:
## Summary

- add a commented-out `RewriteOptions Inherit` option to both virtual host templates
- explain that inherited main-server rules run after rules defined in the virtual host
- keep rewrite inheritance disabled by default

Closes #337.

## Verification

- `./test/build/test_userbuild.sh`
- Apache 2.4.66 `httpd -t` with `RewriteOptions Inherit` in virtual-host context
- comment-stripped effective configuration matches `main` for both templates
- `git diff --check`

### DIFF
--- a/vhosts/templates/example.com.conf
+++ b/vhosts/templates/example.com.conf
@@ -27,6 +27,10 @@
     # Include the basic h5bp config set
     Include h5bp/basic.conf
 
+    # Uncomment to inherit rewrite rules from the main server configuration.
+    # Inherited rules are applied after rules defined in this virtual host.
+    # RewriteOptions Inherit
+
     <Directory "/var/www/example.com/public">
         Require all granted
     </Directory>
--- a/vhosts/templates/no-ssl.example.com.conf
+++ b/vhosts/templates/no-ssl.example.com.conf
@@ -23,6 +23,10 @@
     # Include the basic h5bp config set
     Include h5bp/basic.conf
 
+    # Uncomment to inherit rewrite rules from the main server configuration.
+    # Inherited rules are applied after rules defined in this virtual host.
+    # RewriteOptions Inherit
+
     <Directory "/var/www/example.com/public">
         Require all granted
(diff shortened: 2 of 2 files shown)


LeoColomb (approved):

LeoColomb:
Thanks for submitting this pull request, @oiahoon.

